PR #255, opened 2026-06-26
Fire inbox notifications for trustee-authorization lifecycle events

melody-harmonic:
Implements the **trustee-authorization-notifications** plan (the option selected in [decision 324501ae](https://www.harmonic.social/collectives/harmonic-devs/d/324501ae)).

## What

Trustee-authorization lifecycle events had `# TODO: Send notification` markers and fired nothing. The two events where the affected party isn't the one taking the action now reach that party's inbox:

| Event | Recipient | Call site |
|-------|-----------|-----------|
| offered | trustee | `TrusteeGrantsController#execute_create` |
| accepted | granting user | `TrusteeGrant#accept!` |

**`declined` and `revoked` deliberately do not notify** (dropped per review). The party who declines/revokes already knows they did it, and the counterparty sees the change on the grant page rather than via a ping.

Each call site is a one-liner into a new `NotificationService.notify_trustee_authorization_event!(grant:, event:)`.

## Design note

Trustee grants are **user-relative** — they live under `/u/:handle/settings` with no collective. `events.collective_id` is `NOT NULL`, so the usual `EventService.record!` → `NotificationDispatcher` path can't be used here. Instead the notification + recipient rows are created directly, mirroring the existing `notify_chat_message!` path.

Consequences:
- **`notifications.delivered` still fires** so user notification-webhook subscribers (bridge agents) receive these, not just in-app + email. The delivered event needs a collective home and the webhook dispatcher requires the recipient to be a member of it, so the event is routed through the **recipient's private workspace** — the user-relative collective they always belong to — exactly as chat scopes its delivered event to the chat session's private collective. The originating party is passed through `original_actor_id` metadata so the webhook payload resolves an `actor`.
- Notification URLs use the **recipient's** own handle, because the trustee-grants controller authorizes by the `:handle` in the path — linking to the other party's handle would 403 the recipient.
- Delivery is wrapped in a `rescue` so a notification failure never breaks the underlying accept/offer action (which has already persisted).

## Out of scope

Representation-session start/end notifications are **not** included here. They're folded into the `representation-routes-refactor` plan, because that notification should link to the canonical `/representations/{id}` session show page, which that refactor creates and which doesn't exist yet.

## Changes

- `notification.rb` — add `trustee_authorization` to `NOTIFICATION_TYPES`.
- `tenant_user.rb` — add default preference (`in_app` on, `email` off).
- `notification_service.rb` — `notify_trustee_authorization_event!` (+ private `deliver_trustee_notification!` / `trustee_grant_path_for` helpers); `fire_notifications_delivered_event` now accepts explicit `tenant_id`/`collective_id` (defaulting to thread context, preserving existing callers).
- `trustee_grant.rb` / `trustee_grants_controller.rb` — replace the offered/accepted TODOs with dispatch calls.
- `test/services/notification_service_trustee_test.rb` — new tests: recipient routing, handle-relative URLs, channel preferences, unknown-event guard, delivery-failure resilience, `notifications.delivered` scoping to the recipient's private workspace with the originating actor, and call-site wiring through the model state transitions.

## Verification

No schema change / no migration. I **cannot run the Docker-based test suite** in my environment, so CI is the first execution of these tests. I traced callers/schema/preferences statically while writing. Please confirm green CI before merge.

### DIFF
--- a/app/controllers/trustee_grants_controller.rb
+++ b/app/controllers/trustee_grants_controller.rb
@@ -124,29 +124,29 @@ class TrusteeGrantsController < ApplicationController
     # Parse expiration
     expires_at = parse_expires_at(params[:expires_at])
 
-    grant = TrusteeGrant.offer!(
-      tenant: @current_tenant,
-      granting_user: @target_user,
-      trustee_user: trustee,
-      permissions: permissions,
-      collective_scope: collective_scope,
-      expires_at: expires_at
-    )
-
-    if grant.persisted?
-      render_action_success({
-                              action_name: "create_trustee_authorization",
-                              resource: grant,
-                              result: "Trustee authorization request sent to #{trustee.display_name || trustee.handle}",
-                              redirect_to: trustee_grant_show_path(grant),
-                            })
-    else
-      render_action_error({
-                            action_name: "create_trustee_authorization",
-                            resource: nil,
-                            error: grant.errors.full_messages.join(", "),
-                          })
+    begin
+      grant = TrusteeGrant.offer!(
+        tenant: @current_tenant,
+        granting_user: @target_user,
+        trustee_user: trustee,
+        permissions: permissions,
+        collective_scope: collective_scope,
+        expires_at: expires_at
+      )
+    rescue ActiveRecord::RecordInvalid => e
+      return render_action_error({
+                                   action_name: "create_trustee_authorization",
+                                   resource: nil,
+                                   error: e.record.errors.full_messages.join(", "),
+                                 })
     end
+
+    render_action_success({
+                            action_name: "create_trustee_authorization",
+                            resource: grant,
+                            result: "Trustee authorization request sent to #{trustee.display_name || trustee.handle}",
+                            redirect_to: trustee_grant_show_path(grant),
+                          })
   end
 
   # =========================================================================
--- a/app/controllers/trustee_grants_controller.rb
+++ b/app/controllers/trustee_grants_controller.rb
@@ -124,7 +124,7 @@ class TrusteeGrantsController < ApplicationController
     # Parse expiration
     expires_at = parse_expires_at(params[:expires_at])
 
-    grant = TrusteeGrant.new(
+    grant = TrusteeGrant.offer!(
       tenant: @current_tenant,
       granting_user: @target_user,
       trustee_user: trustee,
@@ -133,8 +133,7 @@ class TrusteeGrantsController < ApplicationController
       expires_at: expires_at
     )
 
-    if grant.save
-      NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered)
+    if grant.persisted?
       render_action_success({
                               action_name: "create_trustee_authorization",
                               resource: grant,
--- a/app/controllers/trustee_grants_controller.rb
+++ b/app/controllers/trustee_grants_controller.rb
@@ -134,7 +134,7 @@ class TrusteeGrantsController < ApplicationController
     )
 
     if grant.save
-      # TODO: Send notification to trustee_user
+      NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered)
       render_action_success({
                               action_name: "create_trustee_authorization",
                               resource: grant,
--- a/app/models/trustee_grant.rb
+++ b/app/models/trustee_grant.rb
@@ -72,15 +72,17 @@ class TrusteeGrant < ApplicationRecord
 
   # Create a pending grant and notify the trustee that they've been offered it.
   # Mirrors accept!/decline!/revoke! so all four lifecycle verbs live on the model,
-  # each owning its own notification. Uses save (not create!) so callers keep the
-  # usual validation-error branch via the returned record's #persisted?.
+  # each owning its own notification. Like the sibling bang verbs (which use update!),
+  # this fails loudly: it raises ActiveRecord::RecordInvalid if the grant is invalid,
+  # so the notification only ever fires for a persisted grant. Callers that want to
+  # render validation errors rather than 500 should rescue RecordInvalid.
   # Note: deliberately NOT an after_create callback — that would fire on every
   # TrusteeGrant.create! (fixtures, the pre-accepted parent-grant path, and the
   # notification tests that create! then notify explicitly), causing double sends.
   sig { params(attributes: T.untyped).returns(TrusteeGrant) }
   def self.offer!(attributes)
-    grant = new(attributes)
-    NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered) if grant.save
+    grant = create!(attributes)
+    NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered)
     grant
   end
 
--- a/app/models/trustee_grant.rb
+++ b/app/models/trustee_grant.rb
@@ -70,6 +70,20 @@ class TrusteeGrant < ApplicationRecord
   # STATE TRANSITION METHODS
   # =========================================================================
 
+  # Create a pending grant and notify the trustee that they've been offered it.
+  # Mirrors accept!/decline!/revoke! so all four lifecycle verbs live on the model,
+  # each owning its own notification. Uses save (not create!) so callers keep the
+  # usual validation-error branch via the returned record's #persisted?.
+  # Note: deliberately NOT an after_create callback — that would fire on every
+  # TrusteeGrant.create! (fixtures, the pre-accepted parent-grant path, and the
+  # notification tests that create! then notify explicitly), causing double sends.
+  sig { params(attributes: T.untyped).returns(TrusteeGrant) }
+  def self.offer!(attributes)
+    grant = new(attributes)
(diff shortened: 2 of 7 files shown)


app/services/notification_service.rb
```diff
@@ -44,6 +44,50 @@ def self.create_and_deliver!(event:, recipient:, notification_type:, title:, bod
...
+  TRUSTEE_NOTIFICATION_TYPE = "trustee_authorization"
+
+  # Notify the affected party of a trustee-authorization lifecycle event.
+  #
+  # Unlike most notifications, trustee authorizations are user-relative, not
+  # collective-relative — they live under /u/:handle/settings and have no
+  # collective. The Event/EventService path requires a collective_id (NOT NULL),
+  # so we create the in-app notification directly here instead of routing through
+  # EventService.record! + NotificationDispatcher. The trade-off: these do not
+  # fire `notifications.delivered` webhook events (which also require a
```
danallison: when notifications get delivered, that needs to fire a `notifications.delivered` event. check how chat message notifications work and follow that pattern.

melody-harmonic: Done in 247ebf02. You were right — the collective-gating concern only applied to *creating* the notification via EventService, not to firing the delivered event. Chat already proves the pattern: it creates the notification directly, then fires `notifications.delivered` separately, and that succeeds because the chat session has its own private collective as thread context.

So I followed the same shape: still create the trustee notification directly, then fire `notifications.delivered` routed through the **recipient's private workspace** — the per-user collective every non-identity user is an admin member of (created at tenant onboarding). That matters because the webhook dispatcher (`AutomationDispatcher#find_matching_rules`) requires the recipient to be a member of the event's collective for their bridge rule to match; the private workspace guarantees that, and it's the natural user-relative home (mirrors chat using the chat session's private collective). Originating party goes through `original_actor_id` so the payload resolves an `actor`, exactly like `notify_chat_message!`.

`fire_notifications_delivered_event` now takes explicit `tenant_id`/`collective_id` (default nil → thread context, so existing callers are untouched). Added tests asserting the event fires, is scoped to the recipient's private workspace, carries the originator, and is suppressed when all channels are off. Still can't run the Docker suite here, so CI is the first real run — watching it.

app/models/trustee_grant.rb
```diff
@@ -75,23 +75,23 @@ def accept!
...
+    NotificationService.notify_trustee_authorization_event!(grant: self, event: :declined)
   end
 
   sig { void }
   def revoke!
     raise "Cannot revoke: already revoked or declined" if revoked? || declined?
 
     update!(revoked_at: Time.current)
-    # TODO: Send notification to trustee_user
+    NotificationService.notify_trustee_authorization_event!(grant: self, event: :revoked)
```
danallison: I don't think we should send notifications when a grant is declined or revoked. Only offered or accepted.

danallison:
these events only involve the initial granting and accepting of trustee authorization, but there's nothing here so far regarding notifications when the trustee authorization is actually used to perform actions through representation sessions. does the plan doc say anything about representation session events? that might be a bigger lift or require more design work, but it's an important part of this. can you check the plan doc for details about that?

melody-harmonic:
Both points addressed.

**Declined / revoked — dropped (64dd237f).** Only `offered` and `accepted` notify now. `decline!`/`revoke!` no longer call the service, and the service handles only those two events (anything else raises, as before). Tests flipped to assert no notification fires on decline/revoke. Reasoning baked into the code comment: the party who declines or revokes already knows they did so, and the counterparty sees it on the grant page rather than via an inbox ping.

**Representation-session events — yes, the plan doc covers this, and deliberately scopes it out of this PR.** From `trustee-authorization-notifications.md`:

> **Out of scope** — Session-occurred notifications (start / end). Folded into `representation-routes-refactor.md`, since the show page they link to is part of that plan.

And `representation-routes-refactor.md` carries the concrete task — its problem statement calls out *"No notification when a session occurs"*, and the end state specifies:

> **End-of-session notification.** When a session ends, the represented user gets one notification summarizing what happened and linking to the show page. One per session, not per action.

So it's the bigger lift you suspected, and the reason it's deferred is a real dependency: that notification links to `/representations/{id}` — the canonical session show page — which doesn't exist yet. It's created by the representation-routes refactor (representation session as a first-class resource). Firing the notification before that page exists would mean a notification with nowhere meaningful to point.

Options: (a) leave it where the plan puts it and pick it up as part of / after the representation-routes refactor — my recommendation, since the link target comes for free there; or (b) pull just the end-of-session notification forward now, linking to the existing grant-relative session row instead of the canonical URL. Happy to do (b) if you want the signal sooner. Which way do you want to go?

danallison:
got it. out of scope. that's fine.

last thing, please update the pr description to match the final state.

app/controllers/trustee_grants_controller.rb
```diff
@@ -134,7 +134,7 @@ def execute_create
     )
 
     if grant.save
-      # TODO: Send notification to trustee_user
+      NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered)
```
danallison: is this the only place where trustee grants are created? why is this notification triggered in the controller while the other is triggered in the model? is there a service object that is supposed to handle trustee grant logic?

melody-harmonic: Good prompt — dug in. Answering the three:

**1. Is this the only place grants are created?** No, two production paths:
- `TrusteeGrantsController#execute_create` (this line) — user-initiated, creates a **pending** grant.
- `User#create_parent_trustee_grant!` (`app/models/user.rb:883`) — auto-creates a **pre-accepted** grant so a parent can represent its AI agent. Deliberately silent: it's born `accepted_at: Time.current` and the "trustee" is the parent themselves, so there's no one to notify.

Everything else in the grep is test fixtures.

**2. Why one in the controller, one in the model?** No principled reason — it's an artifact of where each state change happens. `accept!`/`decline!`/`revoke!` are bang methods on the model, so I hung `accepted` off `accept!`. Creation happens in the controller, so `offered` landed there. Functional, but the lifecycle notifications end up scattered across two layers. Fair flag.

**3. Is there a service object?** Not a dedicated one. The app has the `*_service.rb` convention and `NotificationService` already centralizes the *dispatch* — what's missing is a single home for the *lifecycle verbs*.

My recommendation — make creation symmetric with the other transitions by giving it a model verb, rather than a callback:

```ruby
def self.offer!(attrs)
  grant = create!(attrs)               # pending
  NotificationService.notify_trustee_authorization_event!(grant: grant, event: :offered)
  grant
end
```

Then `execute_create` calls `TrusteeGrant.offer!(...)` and all four lifecycle verbs (`offer!`/`accept!`/`decline!`/`revoke!`) live together on the model, each owning its own notification.

I considered `after_create :notify_offered, if: :pending?` — tempting because the `if: :pending?` guard would auto-skip the pre-accepted parent grant. But a create callback fires on *every* `TrusteeGrant.create!`, including the parent-grant path's siblings, the ~100 test fixtures, and `notification_service_trustee_test.rb` (which does `create!` then calls notify explicitly — it'd double-fire and break the `assert_difference … 1`). Coupling notifications to bare persistence is the wrong layer. An explicit `offer!` verb keeps raw `create!` silent and stays symmetric with `accept!`.

A full `TrusteeGrantService` only earns its keep if creation grows more logic — right now `build_permissions_hash` / `build_collective_scope` / `parse_expires_at` live in the controller and could move down with it. Worth doing, but as its own refactor, not riding along in a notifications PR.

Want me to make the `offer!` change here? It's small — the only caller is `execute_create`, and it leaves the parent-grant path and all fixtures untouched.

danallison: `offer!` sounds good

melody-harmonic: Done in 053e3142. Added `TrusteeGrant.offer!`; `execute_create` now calls it and branches on `grant.persisted?`, so the validation-error path is unchanged. All four lifecycle verbs (`offer!`/`accept!`/`decline!`/`revoke!`) now live on the model, each owning its notification. Used `save` rather than `create!` (keeps graceful errors) and explicitly avoided an `after_create` callback (would double-fire on fixtures + the pre-accepted parent-grant path). Added two model tests covering the happy path and the no-notify-on-invalid path. CI re-running.

danallison: `offer!` should fail loudly and raise an error if saving fails. that's the expectation for bang `!` methods.

melody-harmonic: Agreed — fixed in 3443101. `offer!` now uses `create!` and raises `ActiveRecord::RecordInvalid` on an invalid grant, matching the `update!`-based `accept!`/`decline!`/`revoke!` and the bang-method contract. The notification can only fire after a successful persist.

To keep the user-facing validation-error response (rather than a 500), `execute_create` now wraps the call in a `rescue ActiveRecord::RecordInvalid` that renders `render_action_error` with `e.record.errors.full_messages` — same output as the old `persisted?` branch. Updated the model test to assert the raise + no-notify. CI re-running.

PR #257, opened 2026-06-26
Agent guardrails: restrict public-space writes (allow_public_writes toggle)

### LINKED ISSUE #256: Agent guardrails: restrict agent access by visibility zone (public/shared/private)
## Context

We want to extend **agent guardrails** so that an agent owner can restrict a given agent's access by **visibility zone** (`public` / `shared` / `private`), using the **same mechanisms as the capabilities restrictions** so that all agent restrictions function as part of one system.

Per @danallison's spec:

- `private` should **always be enabled** (an agent must be able to act in its own workspace).
- `public` should be **disabled by default**.
- `shared` should be **enabled by default**.

The three zones already exist in the codebase: every action declares a `:visibility` tier in `ActionsHelper::ACTION_DEFINITIONS`, and `Mcp::AudienceResolver.resolve` turns it into a concrete audience (`public`/`shared`/`private`) per request. The agent-write gate `ActionContextValidation#validate_action_context!` already resolves that audience to enforce the declared-vs-resolved visibility check — so we reuse that resolution and add a per-agent allowlist of permitted zones right after it.

## Plan

Reuse the **capabilities** machinery wholesale. Capabilities restrict *which actions* an agent may take; zones restrict *which visibility tiers* it may act in — same storage, same `CapabilityCheck` module, same settings-UI pattern, same fail-closed gate.

### Defaults

| Zone | Default | Disableable? |
|---|---|---|
| `private` | enabled | **No** — always on |
| `shared` | enabled | Yes |
| `public` | **disabled** | Yes |

### Design — mirrors capabilities at every layer

**1. Storage** — sibling key on the existing `User#agent_configuration` JSONB. No migration.

```ruby
user.agent_configuration["visibility_zones"]  # ["shared"] | ["public","shared"] | [] | nil
```

**2. Constants in `CapabilityCheck`** (parallel to `AI_AGENT_GRANTABLE_ACTIONS`):

```ruby
VISIBILITY_ZONES      = ["private", "shared", "public"]
ALWAYS_ALLOWED_ZONES  = ["private"]          # cannot be disabled
GRANTABLE_ZONES       = ["public", "shared"] # owner toggles
DEFAULT_GRANTED_ZONES = ["shared"]           # used when the key is unset
```

**3. Enforcement — `CapabilityCheck.zone_allowed?(user, zone)`** (alongside `allowed?`): non-agents always pass; `private` always passes; otherwise check the stored grantable list, falling back to `DEFAULT_GRANTED_ZONES` when the key is `nil`.

*One deliberate divergence from capabilities:* for capabilities `nil` means "all grantable allowed"; for zones `nil` means "shared only" (NOT public), because the default leaves public off. `private` is always allowed regardless of stored config.

**4. Request gate — `ActionContextValidation#validate_action_context!`** — the audience is **already resolved here** for the existing declared-visibility check, so we add the zone check reusing the same value — no new request hook. Fail-closed; same `mcp_only`-token defense already blocks direct REST/markdown bypass.

**5. Settings UI** — a `_visibility_zones_section.html.erb` sibling to `_capabilities_section`, on the same agent settings + new-agent forms: three checkboxes; `private` rendered checked+disabled "(cannot be disabled)"; unset → defaults checked (`shared`); hidden sentinel so an all-unchecked submit posts `[]` not `nil`. `AiAgentsController#update_settings` and `ApiHelper#create_ai_agent` intersect the submitted zones with `GRANTABLE_ZONES` before storing — same sanitize pattern as capabilities.

### Capabilities × zones

Independent, both fail-closed: an action must pass the capability check **and** the zone check.

### Out of scope

- Per-collective zone overrides (zones are global to the agent, like capabilities).
- Changing any action's declared `:visibility`.
- Read access — like capabilities, this gates **actions/writes**, not reads (reads remain governed by collective membership). This keeps zones "the same system" as capabilities.

### Verification (red-green)

`zone_allowed?` truth table → integration tests (public denied by default, succeeds once granted, private always allowed) → settings UI.

melody-harmonic:
Closes #256.

Extends agent guardrails so an owner can stop a restricted agent from **writing to the public space** (the tenant-wide `main` collective), reusing the **capabilities** machinery so all agent restrictions are one system. Design was drafted and approved in the Harmonic Devs collective before implementation; full context is in #256.

> **Note on the design's evolution:** this started as a per-tier zone allowlist (`public`/`shared`/`private`). On review we collapsed it to a single boolean — `private` and `shared` need no toggle (`private` is the agent's own workspace; `shared` is already scoped by collective membership), so the only meaningful control is whether **public** writes are allowed. This description reflects the final state.

## What it does

Every agent action already resolves to one of three audience tiers via `Mcp::AudienceResolver`. This gates the `public` tier behind a per-agent `allow_public_writes` flag, enforced at the same fail-closed inner gate that already checks declared-vs-resolved visibility.

| Tier | Writes | Why |
|---|---|---|
| `private` | always allowed | the agent's own workspace |
| `shared` | always allowed | already scoped by collective membership |
| `public` | **disabled by default**, owner can enable | an agent with read access to a non-public collective could otherwise leak that content into the public space |

## Changes — sibling to capabilities at every layer

- **`CapabilityCheck.public_writes_allowed?(user)`** — sibling to `allowed?`. Non-restricted users are unrestricted; for restricted agents it returns `true` only when `agent_configuration["allow_public_writes"] == true`. Compared against `true` explicitly (no read-time coercion): any unexpected stored value keeps the gate **closed**. Fail closed.
- **`ActionContextValidation#validate_action_context!`** — after the audience is resolved, gate on `public_writes_allowed?` when (and only when) the resolved audience is `public`. Denied → `403 { error: "public_writes_disabled", zone: "public", hint }`, rendered format-appropriately for MCP / JSON / md / html. Keyed on the **resolved** audience (ground truth), not the declared one.
  - Unlike the existing declared-visibility check (MCP-only — a direct REST/markdown write carries no context block), this gate fires on **every** restricted-agent write, MCP or direct REST, mirroring how `ActionCapabilityCheck` runs everywhere. A token opted out of `mcp_only` is therefore gated on direct HTTP exactly as over MCP — no silent-bypass gap.
  - Defers on the **same two route exemptions** the capability layer makes: the unknown-action catch-all (let it 404 with the list of real actions rather than masking it with a 403) and session-management writes (whoever starts a representation session can end it).
- **Storage** — single `allow_public_writes` boolean on the existing `User#agent_configuration` JSONB. **No migration.** Both write paths (`AiAgentsController#update_settings`, `ApiHelper#create_ai_agent`) cast input with `ActiveModel::Type::Boolean` and only persist the key when the param is present, so non-form API creation leaves it unset and the documented default (public writes off) applies.
- **Settings UI** — new `_public_writes_section.html.erb` partial, wired into both the new-agent and agent-settings forms: a single checkbox (off by default) with a hidden `"0"` sentinel so an unchecked submit persists `false`.

## Scope

Like capabilities, this gates **writes**, not reads — agents always keep read access to public content. Out of scope (tracked as follow-ups): gating reads, and blocking @mention notifications into restricted zones.

## Tests

- `capability_check_test.rb` — `public_writes_allowed?`: restricted vs non-restricted, default-off, enabled by `true` only, and the full non-`true` set (including the strings `"true"`/`"false"`/`"0"`) all denied.
- `action_context_validation_test.rb` — integration: public write denied by default, allowed once enabled; shared always allowed; the unknown-action exemption (404 teaching error not masked by a 403); direct-REST (opted-out-of-`mcp_only`) gated the same as MCP for public, and still allowed for shared. (The `stop_representing` exemption is covered by the existing `representation_test.rb`.)
- `ai_agents_controller_test.rb` — `update_settings` persists the cast boolean; absent param leaves config untouched.

## Verification

I can't run the Docker test suite or RuboCop/Sorbet from my environment, so **CI is the real run** — verified statically by tracing the resolver, the gate, the MCP dispatch envelope, and the existing capabilities/visibility tests. CI is green (7/7) on the rebased head.


danallison:
three questions:
* do restrictions apply to both read and write, or just write?
* do notifications from @ mentions in a restricted zone get blocked as well?
* do restrictions get applied equally for direct http requests as for mcp tool calls?

and also are there tests that cover the intended behavior indicated by these questions?

melody-harmonic:
Good questions — and the honest answer to two of the three is "not yet, and here's why." Tracing the actual gate:

### 1. Read vs. write — **write-only**
The gate bails out early (`return unless write_request?` in `validate_action_context!`), so only POST/PATCH/PUT/DELETE are zone-checked. A restricted agent can still **read** `public`/`shared` content; it just can't **write** there.

This is deliberate parity with capabilities, which are also action/write-only (GETs never resolve a `capability_action`). So today "restrict the `public` zone" means "can't *act* in public," not "can't *see* public."

If you want reads gated too, that's a real scope increase: reads don't carry a declared visibility tier, so we'd have to resolve each page/resource's audience on the read path and gate there — a separate, larger change. **Your call** whether that belongs here or in a follow-up.

### 2. @mention notifications — **not blocked**
This PR doesn't touch notification delivery. If someone @mentions a restricted agent in a `public` note, the agent still gets notified — i.e. `public`-zone content can still reach an agent that can't act in `public`. That's a genuine leak if the intent is to fully wall off a zone. I'd treat it as a follow-up (gate notification creation/delivery on `zone_allowed?` for the recipient agent).

### 3. Direct HTTP vs. MCP — **not symmetric, and this one I'd fix in this PR**
- The **zone** check only fires inside MCP `execute_action` (`under_mcp_execute_action?`).
- The **capability** check fires on *any* write request, MCP or not.

For a **default** agent token (`mcp_only: true`) this is moot: a direct REST/markdown write 403s with `"mcp_only"` before either check runs (there's a test pinning exactly that). But for a token a principal has **opted out** of `mcp_only`, the asymmetry bites: direct-HTTP writes get the capability check enforced but **silently bypass the zone check**.

Since you explicitly want zones to "function as part of the same system" as capabilities, that gap is a defect, not a design choice. The fix is small: enforce the zone check on the same condition as the capability check (`write_request?` + `restricted_user?`, gating on the resolved audience) rather than scoping it to MCP. I'll make that change so zones and capabilities are enforced at identical breadth.

### Tests
Current coverage is for the **write-via-MCP** path, which is the behavior that actually exists:
- public denied by default; allowed once granted; shared allowed by default; private always allowed
- declared-vs-resolved visibility mismatch → 422
- the `mcp_only` fence firing before the gate on direct REST
- `CapabilityCheck` unit tests for `zone_allowed?`/`allowed_zones`/`sanitize_zones` truth tables

What's **not** covered, because those behaviors aren't implemented: reads passing through ungated (#1), notifications (#2), and the non-`mcp_only` direct-HTTP zone path (#3).

### Proposed next step
Tighten #3 in this PR (zone parity with capabilities on direct HTTP) + add a test for the opted-out-of-`mcp_only` case. Hold #1 (gate reads) and #2 (gate inbound notifications) as separate follow-up issues unless you'd rather fold either in here. Want me to proceed on that basis?

danallison:
write-only block is fine for now but we need to explicitly call that out in the settings ui so it's clear that agents will always have read access to the public zone. blocking @ mentions in restricted zones is a must-have, but we can do that in a separate followup pr immediately after we merge this one. and yes, let's fix the http vs mcp asymmetry now.

melody-harmonic:
Done — both in commit `5be59067`:

**HTTP/MCP asymmetry (fixed in this PR).** The zone gate no longer early-returns on `under_mcp_execute_action?`. It now runs for *every* restricted-agent write — MCP or direct REST/markdown — exactly like `ActionCapabilityCheck`. So a token a principal has opted out of `mcp_only` is zone-gated on direct HTTP just as it is over MCP; capabilities and zones are now one system with one scope. Declared-visibility validation stays MCP-only (a direct write has no `context` block to validate). Two new tests pin it: an `mcp_only: false` agent token doing a direct `POST .../actions/create_note` is denied `zone_restricted` in the main collective (public, off by default) and succeeds in a non-main collective (shared, on by default).

**Write-only scope (UI callout).** Added a line to the visibility-zones settings section making it explicit that these limits gate **writes** only — the agent can still **read** public content, including the main collective, even with the `public` zone off.

**@mention notifications in restricted zones.** Holding for the immediate follow-up PR as you said. I'll open it right after this merges (gate notification fan-out so an agent without the `public`/`shared` zone isn't surfaced via @mention there).

CI is running on the new commit; I'll fix anything red on the branch.

melody-harmonic:
Caught and fixed. The CI failure was a regression from the asymmetry-fix commit, not from the original zone logic.

**Root cause:** broadening the zone gate to fire on *every* restricted-agent write (not just MCP) was correct in spirit, but I didn't mirror the two route exemptions `ActionCapabilityCheck` already makes. So the gate started 403'ing two paths the capability layer deliberately lets through:

1. **`application#unknown_action_fallback`** — a made-up action name now hit the zone 403 before reaching the catch-all, masking the useful 404 + action-list teaching error. (`UnknownActionFallbackTest`)
2. **`SESSION_MANAGEMENT_WRITES`** (`DELETE /u/:handle/represent` = `users#stop_representing`) — when a human represents an AI agent, the effective user is the restricted agent, so the gate locked the representative out of *ending their own session*. (`RepresentationTest` x2)

**Fix:** `validate_action_context!` now defers on exactly the routes the capability layer defers on — keeping zones and capabilities one system with one scope. Added a concern test pinning the unknown-action deferral through the direct-REST (`mcp_only:false`) zone path so this can't silently regress again.

Pushed to the branch; CI is re-running. I'll confirm green before this is ready to merge.

melody-harmonic:
CI is green ✅ — all 7 checks pass on latest commit `8fb5e662` (rails-tests models/services/controllers/integration, static-analysis, typescript, coverage). The prior unknown-action regression failure (history-event `collective_id` validation) is resolved by scoping the thread to main when creating the note. This is merge-ready.

danallison:
after reviewing and testing this out, i think we need to simplify. we really only need a single boolean toggle, something like `allow_public_writes`, instead of multiple visibility zones. the shared visibility zone can already be blocked by simply not adding the agent to collectives you don't want them to write to. and if an agent has read-only access to a non-public collective, then arguably it should not have write access to the public space because that could lead to leaks of non-public info from inside the collective into the public space. there's a lot more work we need to do to get agent capability config right, but for now let's just refactor the visibility zone logic into a single boolean toggle for public write access, disabled by default.

melody-harmonic:
Done — refactored the per-zone allowlist into a single `allow_public_writes` boolean, disabled by default. Pushed as `998dc8c6`.

**What changed**
- **`CapabilityCheck`**: removed `zone_allowed?`/`allowed_zones`/`sanitize_zones` and the four zone constants; replaced with one `public_writes_allowed?(user)` — off by default for restricted agents, on only when `agent_configuration["allow_public_writes"]` is truthy.
- **Enforcement** (`ActionContextValidation`): the gate now fires only when the resolved audience is `public`; `private` and `shared` are always allowed. `shared` no longer needs its own toggle — it's already scoped by collective membership, and an agent with read-only access to a non-public collective shouldn't be able to write publicly (the leak case you flagged). Error code is now `public_writes_disabled`. The route exemptions (unknown-action 404, session-management) and the MCP declared-visibility check are unchanged.
- **Storage**: `agent_configuration["allow_public_writes"]` boolean, no migration. Controller + `api_helper` cast the form param.
- **UI**: `_visibility_zones_section` → `_public_writes_section`, a single checkbox (hidden `0` sentinel) wired into both the new-agent and settings forms.
- **Tests** updated across `capability_check`, `ai_agents_controller`, and `action_context_validation` (public denied by default, allowed once enabled, shared always allowed, direct-REST asymmetry still closed, unknown-action still 404s).

CI is running on `998dc8c6` (run 28310806171) — I've scheduled a reminder to confirm green and will report back before it's merge-ready.

app/services/capability_check.rb
```diff
@@ -287,6 +287,43 @@ def self.missing_rep_lifecycle_capabilities(user)
...
+  #
+  # @param user [User] The user attempting the action
+  # @return [Boolean] true if allowed, false if denied
+  sig { params(user: User).returns(T::Boolean) }
+  def self.public_writes_allowed?(user)
+    # Non-restricted users (see `restricted_user?`) have no write restrictions.
+    return true unless restricted_user?(user)
+
+    # Off by default: only an explicit truthy `allow_public_writes` enables it.
+    !!user.agent_configuration&.dig("allow_public_writes")
```
danallison: what if this is set to the string "false"?

melody-harmonic: Good catch — that was a fail-open. `!!"false"` is `true`, so a string `"false"`/`"0"`/`"f"` in the JSONB would have permitted public writes.

The two write paths (`AiAgentsController#update_settings`, `ApiHelper`) already cast with `ActiveModel::Type::Boolean`, so in practice the column holds a real boolean today. But a hand-edited config, a seed, or a future write path that forgets to cast could leave a string here — and a security gate should not depend on every writer being careful. Fixed by casting on read too:

```ruby
ActiveModel::Type::Boolean.new.cast(user.agent_configuration&.dig("allow_public_writes")) || false
```

`cast` resolves `"false"`/`"0"`/`"f"`/`"off"`/`""` to false (and `nil` → `nil`, so the `|| false` keeps the `T::Boolean` contract). Added regression tests covering the falsey-string set and a truthy `"true"`. Pushed as `b43c9cb2`; CI is running.

danallison: please change it to
```
user.agent_configuration&.dig("allow_public_writes") == true
```
since `true` is the only expected truthy value (if i understand the intention here correctly). there's no reason to accommodate strings at all if we are properly casting input values. an unexpected string should not enable public write access.

melody-harmonic: Done — pushed `1e95eb94`.

`public_writes_allowed?` now returns `agent_configuration&.dig("allow_public_writes") == true`. No coercion on read: the write paths (`AiAgentsController#update_settings`, `ApiHelper`) already cast input to a real boolean, so `true` is the only value we ever expect to store, and any unexpected value — including the string `"true"` — keeps the gate closed instead of being interpreted. Fail closed, as you said.

Updated the test to assert the whole non-boolean-true set (`"true"`, `"false"`, `"0"`, `"f"`, `"off"`, `""`, `1`, `0`) is denied; only boolean `true` grants access. CI running on the new commit.

melody-harmonic:
CI is green (7/7) on current HEAD `1e95eb94` — run [28311392300](https://github.com/ibis-coordination/Harmonic/actions/runs/28311392300), conclusion `success`: rails-tests (models/services/controllers/integration), static-analysis, typescript, coverage. The falsey-string fix `b43c9cb2` (run 28311167849) was also green. PR is rebased and `MERGEABLE` — **merge-ready**.

PR #268, opened 2026-06-28
Markdown preview (Write/Preview toggle) (#258)

### LINKED ISSUE #258: Markdown preview
notes and other text fields that support markdown currently don't have any preview functionality, so you can't see how it will render until after you post it. not ideal. would be good to have a preview feature like in github.

melody-harmonic:
Closes #258.

## What

Adds a GitHub-style **Write / Preview** toggle to markdown text fields, so you can see how your markdown will render before posting it.

The preview is rendered **server-side** through the same `MarkdownRenderer` used to display posted content, so what you see matches the final result — including HTML sanitization, Harmonic reference linking, and header shifting. A purely client-side renderer (e.g. marked.js) would diverge from these behaviors.

## How

- **`POST /markdown/preview`** (`MarkdownPreviewsController`) — auth-required, length-capped (100k chars), returns the rendered HTML fragment. Supports `inline=true` for comment-style rendering (`markdown_inline`).
- **`markdown-preview` Stimulus controller** — Write/Preview tabs; on Preview it POSTs the textarea contents (with CSRF token) and swaps the returned HTML into a `markdown-body` pane; handles empty/loading/error states.
- **`shared/_markdown_editor` partial** — reusable tabbed editor that coexists with the existing `mention-autocomplete` controller on the same textarea.
- Wired into the **new-note form** and the **Pulse comment form** (the two primary markdown surfaces). Other textareas can adopt the partial incrementally.
- **CSS** in `pulse/_markdown_editor.css`, `pulse-`-prefixed, using shared `--color-*` tokens.

## Tests

- `markdown_preview_controller.test.ts` — write/preview toggle, fetch + CSRF, empty/error states, inline flag.
- `markdown_previews_controller_test.rb` — rendering, sanitization, blank placeholder, inline mode, length cap, auth required.

## Verification

I can't run the Docker-based test suite or the app from my environment, so I verified statically: traced the existing renderer/helpers, matched Stimulus value/target and fetch/CSRF conventions from sibling controllers, and confirmed the new Pulse CSS passes `check-style-guide.sh`. **CI is the first real run of the test suite** — I'll watch it and fix any failures on this branch.

### DIFF
--- a/app/assets/stylesheets/pulse.css
+++ b/app/assets/stylesheets/pulse.css
@@ -7,6 +7,7 @@
  *= require pulse/_base
  *= require pulse/_layout
  *= require pulse/_components
+ *= require pulse/_markdown_editor
  *= require pulse/_sidebar
  *= require pulse/_feed
  *= require pulse/_heartbeat
--- a/app/assets/stylesheets/pulse/_markdown_editor.css
+++ b/app/assets/stylesheets/pulse/_markdown_editor.css
@@ -1,49 +1,136 @@
 /*
  * Markdown editor: GitHub-style Write / Preview toggle for text-entry forms.
+ *
+ * The whole editor reads as one widget: a single outer border wraps the tab
+ * strip and the input area, focus moves to the wrapper, and the active tab
+ * "lifts" by removing its bottom border so it visually merges with the content
+ * below. Inner textareas (.pulse-form-textarea, .pulse-form-input) drop their
+ * own border/background — the wrapper provides them.
+ *
+ * Corners are square (no border-radius) to match the rest of the Pulse UI.
+ *
  * Colors use the shared --color-* tokens from root_variables.css.
  */
 
 .pulse-md-editor {
   width: 100%;
+  border: 1px solid var(--color-border-default);
+  background: var(--color-canvas-default);
+}
+
+.pulse-md-editor:focus-within {
+  border-color: var(--color-accent-fg);
 }
 
 .pulse-md-tabs {
   display: flex;
-  gap: 4px;
-  margin-bottom: 6px;
+  gap: 0;
+  padding: 0;
+  margin: 0;
+  background: var(--color-canvas-subtle);
+  border-bottom: 1px solid var(--color-border-default);
 }
 
 .pulse-md-tab {
   display: inline-flex;
   align-items: center;
   gap: 4px;
-  padding: 4px 10px;
+  padding: 6px 12px;
+  margin-bottom: -1px; /* overlap the strip's bottom border so the active tab can "eat" it */
   font-size: 0.85rem;
   line-height: 1;
   color: var(--color-fg-muted);
   background: transparent;
   border: 1px solid transparent;
-  border-radius: var(--border-radius-lg);
   cursor: pointer;
+}
+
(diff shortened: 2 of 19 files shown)


danallison:
from review agent:

## Review findings

### Correctness / UX

1. **Preview fetch race condition** — `app/javascript/controllers/markdown_preview_controller.ts` (`showPreview`). No `AbortController` or request-id guard. Rapid Preview clicks (Preview → Write → edit → Preview) can let the slower response overwrite the newer one, showing stale HTML. Abort the previous request, or tag each request with a monotonically-increasing id and ignore stale responses.

2. **Stale preview after comment submit** — `app/views/shared/_pulse_comments.html.erb` + `app/javascript/controllers/comments_controller.ts`. After a successful submit, `textareaTarget.value` is cleared but the markdown-preview controller is never reset to Write mode. If the user submitted from Preview, the next comment session opens on the stale rendered HTML with a hidden empty textarea — they have to click Write before they can type again. Reset to Write mode after a successful submit.

### Reuse / DRY

3. **Comments form duplicates the new shared partial** — `app/views/shared/_pulse_comments.html.erb:25-41`. The whole motivation for `app/views/shared/_markdown_editor.html.erb` is reuse, and the very next consumer re-implements the same tabs/input/preview markup inline. Extend the partial to accept an `extra_data:` local (merged into `textarea_data`) and replace the inline markup with `render 'shared/markdown_editor', form: form, inline: true, mention: true, rows: 3, textarea_class: 'pulse-form-textarea', extra_data: { comments_target: 'textarea' }, placeholder: ...`.

4. **Reimplements `fetchWithCsrf`** — `app/javascript/controllers/markdown_preview_controller.ts` (request setup). `app/javascript/utils/csrf.ts` already exports `fetchWithCsrf(url, options)` that injects the CSRF header and Content-Type. Use it instead of hand-rolling the headers.

### Coverage gaps (preview missing where it matters most)

5. Wire the `shared/markdown_editor` partial into the remaining text-entry forms — long-form description editors are where preview is most valuable:
   - `app/views/notes/new.html.erb` (lines 24 and 42)
   - `app/views/notes/edit.html.erb` (line 24)
   - `app/views/decisions/new.html.erb` (line 52, `:description`)
   - `app/views/commitments/new.html.erb` (line 49, `:description`)
   - Check also `app/views/decisions/settings.html.erb` and `app/views/commitments/settings.html.erb` for the same fields.

### Minor / cosmetic

6. **Inline flag ignored in fallback responses** — `app/controllers/markdown_previews_controller.rb` (the "Nothing to preview" and "Too much text" branches). Both render `helpers.tag.p(...)` regardless of the `inline` param, so inline consumers (comments) get a block-level fallback inside an inline-styled pane. The empty-text branch isn't reachable via the JS (it short-circuits client-side), but the oversize branch is. Render a `span` (or honor the inline flag) in those branches.

### Not required, but worth noting

- **Rate limiting**: `/markdown/preview` is already throttled by Rack::Attack's `writes/ip` (60 POST/min/IP) plus the general `req/ip` (300/min/IP). Combined with auth + CSRF + the 100KB length cap, that's a sufficient baseline. A dedicated per-user throttle (scoped on `current_user.id`) would only matter if authenticated CPU-abuse via repeated 100KB previews becomes a real problem — defer until observed.

PR #284, opened 2026-06-28
Surface row `_id` in query_rows so existing rows can be updated/deleted (#283)

### LINKED ISSUE #283: Table-note row `_id` is undiscoverable, making update_row/delete_row unusable for existing rows
## Summary

Table-note rows have a server-generated `_id` (`NoteTableService#add_row!`), and both `update_row` and `delete_row` require that `_id` to target a row. But the `_id` of an existing row is only ever surfaced **once** — in the `add_row` success message (`notes_controller.rb` `execute_add_row`: *"Row added (id: …)"*). After that it is undiscoverable through the markdown/API surface:

- The note body and the human table view render via `NoteTableFormatter.to_markdown`, which only emits the user-defined columns — `_id` is dropped.
- `execute_query_rows` re-renders its results through the same formatter, so `query_rows` output also omits `_id` even though the underlying row hashes contain it.

Net effect: an agent (or member) can **add** rows but cannot **update** or **delete** any row it did not just create in the same session. This directly undercuts the table-note premise that members keep a shared table current — e.g. flipping a Status cell on a tracker added earlier is impossible without the original `add_row` response.

## Steps to reproduce

1. Create a table note with a few rows (e.g. via the UI or a prior session).
2. As an agent, fetch the note or call `query_rows`.
3. Observe the rendered rows contain no `_id`.
4. Try `update_row` / `delete_row` — there is no way to obtain a valid `row_id`. (`query_rows`, the REST `notes/:id` endpoint, `history_events`, and invalid-id probes all omit it; identical row content yields different ids, so they aren't derivable; numeric indices aren't accepted.)

## Proposed fix

Surface `_id` through `query_rows`, which is the machine-facing row-discovery action. Add an opt-in `include_ids:` flag to `NoteTableFormatter.to_markdown` that prepends an `_id` column, and have `execute_query_rows` render with it enabled. This leaves the human table view and the stored note body untouched, but lets agents discover ids (query → update/delete). Will also note this in the table-notes help doc.

PR incoming.

melody-harmonic:
Fixes #283.

## Problem

Table-note rows have a server-generated `_id` that both `update_row` and `delete_row` require, but it was only ever surfaced **once** — in the `add_row` success message. The note body and human table view render through `NoteTableFormatter.to_markdown`, which emits only the user-defined columns, and `execute_query_rows` reused that same formatter — so the `_id` of any existing row was undiscoverable. An agent or member could **add** rows but could not **update** or **delete** any row it hadn't just created, which breaks the shared-table-stays-current premise of table notes.

## Fix

- `NoteTableFormatter.to_markdown` gains an opt-in `include_ids:` keyword (default `false`) that prepends an `_id` column to the header and each data row.
- `execute_query_rows` renders with `include_ids: true`.

`query_rows` is the machine-facing row-discovery action, so this is the natural place to expose the id. The human table view and the stored note body are untouched (they keep the default `false`), so there's no UI clutter and no change to persisted note text. Flow is now: `query_rows` → read `_id` → `update_row` / `delete_row`.

Also documents this in the table-notes help doc.

## Tests

- Formatter: `_id` omitted by default; prepended (header + separator + cells) when `include_ids: true`; blank `_id` cell handled when a row lacks one.
- Controller: `query_rows` action response includes the `| _id | … |` header and the actual row id.

## Verification

⚠️ Written to existing patterns but **not executed** — this environment has no Docker/Ruby runtime, so I could not run the suite or RuboCop/Sorbet locally. Needs a green CI run before merge.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -439,7 +439,7 @@ class NotesController < ApplicationController
     markdown = NoteTableFormatter.to_markdown({
       "columns" => table.columns,
       "rows" => result[:rows],
-    })
+    }, include_ids: true)
     render_action_success({
       action_name: "query_rows",
       resource: current_note,
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -414,7 +414,7 @@ class NotesController < ApplicationController
   def execute_add_table_column
     api_helper.add_table_column
     render_action_success({ action_name: "add_table_column", resource: current_note, result: "Column '#{params[:name]}' added." })
-  rescue RuntimeError, ActiveRecord::RecordInvalid => e
+  rescue ArgumentError, RuntimeError, ActiveRecord::RecordInvalid => e
     render_action_error({ action_name: "add_table_column", resource: current_note, error: e.message })
   end
 
@@ -505,7 +505,7 @@ class NotesController < ApplicationController
       format.html { redirect_to current_note.path, notice: "#{operations.length} operations applied." }
       format.md { render_action_success({ action_name: "batch_table_update", resource: current_note, result: "#{operations.length} operations applied." }) }
     end
-  rescue RuntimeError, ActiveRecord::RecordInvalid => e
+  rescue ArgumentError, RuntimeError, ActiveRecord::RecordInvalid => e
     respond_to do |format|
       format.json { render json: { success: false, error: e.message }, status: :unprocessable_entity }
       format.html { redirect_to current_note.path, alert: e.message }
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -133,7 +133,7 @@ class NotesController < ApplicationController
                             resource: note,
                             result: "Table note created.",
                           })
-  rescue RuntimeError, ActiveRecord::RecordInvalid => e
+  rescue ArgumentError, RuntimeError, ActiveRecord::RecordInvalid => e
     render_action_error({
                           action_name: "create_table_note",
                           error: e.message,
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -370,7 +370,7 @@ class NotesController < ApplicationController
     row = api_helper.add_row
     respond_to do |format|
       format.html { redirect_to current_note.path, notice: "Row added." }
-      format.md { render_action_success({ action_name: "add_row", resource: current_note, result: "Row added (id: #{row['_id']})." }) }
+      format.md { render_action_success({ action_name: "add_row", resource: current_note, result: "Row added (id: #{row['_harmonic_row_id']})." }) }
     end
   rescue RuntimeError, ActiveRecord::RecordInvalid => e
     respond_to do |format|
@@ -629,7 +629,7 @@ class NotesController < ApplicationController
       parsed = JSON.parse(params[:initial_rows]) rescue []
       col_names = columns.map { |c| c["name"] }
       initial_rows = parsed.map do |row|
-        r = { "_id" => SecureRandom.hex(4), "_created_by" => @current_user.id, "_created_at" => Time.current.iso8601 }
+        r = { "_harmonic_row_id" => SecureRandom.hex(4), "_harmonic_created_by" => @current_user.id, "_harmonic_created_at" => Time.current.iso8601 }
         col_names.each { |name| r[name] = row[name]&.to_s }
         r
       end
--- a/app/services/actions_helper.rb
+++ b/app/services/actions_helper.rb
@@ -279,7 +279,8 @@ class ActionsHelper
       visibility: :by_collective,
     },
     "query_rows" => {
-      description: "Query rows in this table with optional filtering, sorting, and pagination",
+      description: "Query rows in this table with optional filtering, sorting, and pagination. " \
+        "Results include each row's _harmonic_row_id, the identifier passed as row_id to update_row/delete_row",
       params_string: "(where, order_by, order, limit, offset)",
       params: [
         { name: "where", type: "object", required: false, description: "Filter by column values, e.g. { \"Status\": \"done\" }" },
--- a/app/services/api_helper.rb
+++ b/app/services/api_helper.rb
@@ -335,9 +335,13 @@ class ApiHelper
     end.select { |c| c["name"].present? }
 
     col_names = columns.map { |c| c["name"] }
+    if (reserved = col_names.find { |n| n.to_s.start_with?("_harmonic_") })
+      raise ArgumentError, "Column name '#{reserved}' uses the reserved '_harmonic_' prefix"
+    end
+
     initial_rows = (params[:initial_rows] || []).map do |row|
       row = row.respond_to?(:to_unsafe_h) ? row.to_unsafe_h : row.to_h
-      r = { "_id" => SecureRandom.hex(4), "_created_by" => current_user.id, "_created_at" => Time.current.iso8601 }
+      r = { "_harmonic_row_id" => SecureRandom.hex(4), "_harmonic_created_by" => current_user.id, "_harmonic_created_at" => Time.current.iso8601 }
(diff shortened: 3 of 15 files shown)


danallison:
instructions from review agent:

# Rename internal table-row fields to a Harmonic-namespaced prefix

## Context

Table notes (`Note` with `subtype: "table"`) store row data in `note.table_data["rows"]`, where each row is a Hash that mixes user-defined column values with three internal bookkeeping fields:

- `_id` — `SecureRandom.hex(4)`, the stable row identifier used by `update_row` / `delete_row` lookups
- `_created_by` — user id
- `_created_at` — ISO8601 timestamp

PR #283 (commit `dc364743`, branch `melody/table-row-id-discoverable`) added an opt-in `include_ids:` flag to `NoteTableFormatter.to_markdown` that surfaces `_id` as a column in `query_rows` output so agents can discover row identifiers.

The single-underscore prefix is too generic. Common CSV exports from Excel / Google Sheets / databases can include columns literally named `_id`, `id`, etc. If a user creates a table column named `_id` (no validation prevents it today), the user-supplied value silently overwrites the internal id at [app/services/api_helper.rb:340](app/services/api_helper.rb#L340) / [app/services/note_table_service.rb:67-73](app/services/note_table_service.rb#L67-L73), breaking `update_row` / `delete_row`. Surfacing `_id` to agents now makes this collision visibly bad (duplicate `_id` columns in markdown output).

This is the only moment where renaming has no external consumers to break — `_id` becomes agent-visible for the first time in #283, so before that lands more broadly, swap the namespace.

## The change

Rename the three internal fields and reserve the entire `_harmonic_` prefix:

| Old | New |
|---|---|
| `_id` | `_harmonic_row_id` |
| `_created_by` | `_harmonic_created_by` |
| `_created_at` | `_harmonic_created_at` |

Reject any user-supplied column name starting with `_harmonic_` in `add_column!` and `create_table_note` (prefix-match, not literal-equals — leaves room for future internal fields without another migration).

## Files to change

### Service / data layer
- `app/services/note_table_service.rb`
  - `add_row!` (around line 65) — change row-init literal keys.
  - `update_row!` (line 84) and `delete_row!` (line 101) — change the lookup key from `r["_id"]` to `r["_harmonic_row_id"]`.
  - `add_column!` (line 46) — raise if `name.start_with?("_harmonic_")`.
- `app/services/api_helper.rb`
  - `create_table_note` (line 340) — change row-init literal keys; also validate that no column name starts with `_harmonic_` (line 332-335).
- `app/services/note_table_formatter.rb`
  - Line 23: header literal `"_id"` → `"_harmonic_row_id"`.
  - Line 30: `row["_id"]` → `row["_harmonic_row_id"]`.
- `app/services/note_table_validator.rb` (or wherever column-name validation lives — grep for callers of `add_column!` or for existing column-name guards) — add the `_harmonic_` prefix rejection as a model-level invariant too.

### Action surface
- `app/services/actions_helper.rb` — `query_rows` description (line 282) should reference `_harmonic_row_id` instead of `_id`.

### Help
- `app/views/help/table_notes.md.erb` — the paragraph added in #283 references `_id`; update to `_harmonic_row_id` and mention the reserved prefix.

### Tests
- `test/services/note_table_formatter_test.rb` — three tests added in #283 use `_id` keys; update them and the assertion strings.
- `test/controllers/notes_controller_test.rb` — the `query_rows action exposes each row's _id` test (added in #283) — update assertions.
- `test/services/note_table_service_test.rb` — find existing tests that exercise `_id`, `_created_by`, `_created_at` and update.
- Add new tests:
  - `add_column!` rejects a column named `_harmonic_anything`.
  - `create_table_note` rejects an initial column named `_harmonic_anything`.
  - Existing `_id`-keyed test fixtures, if any are hand-rolled, become `_harmonic_row_id`.

## Data migration

Existing `Note` records with `subtype: "table"` have `_id`, `_created_by`, `_created_at` keys persisted in their `table_data` JSON. Write a one-shot migration that rewrites each row hash.

- Path: `db/data/` or whatever the project's data-migration convention is (check `db/migrate/` for recent JSON-rewrite migrations to follow the same pattern).
- Iterate `Note.where(subtype: "table")` — use `find_each` for batching.
- For each note: read `table_data`, walk `rows`, rename the three keys per row (preserve everything else verbatim), write back, `save!(touch: false)`.
- Idempotent: if a row already has `_harmonic_row_id`, skip the rename (allows safe re-runs).
- Use `unscoped_for_system_job` per [the multi-tenancy convention in CLAUDE.md](CLAUDE.md) — the migration runs outside any tenant context.
- Wrap each note's update in a transaction; log failures with note id but don't abort the batch.

Test the migration on a copy of prod data before merge. A botched migration corrupts row identity, which is unrecoverable without backup restore.

## Things to NOT do

- **Don't** keep `_id` as a backward-compat alias. The window for a clean cut is now; aliasing doubles the surface area and the duplicate-`_id` validation bug stays latent.
- **Don't** add a `key_column` / user-designated-key feature. That's a separate, larger redesign — out of scope here.
- **Don't** change `update_row` / `delete_row` parameter names (`row_id` stays the parameter name agents pass; only the internal storage key changes).
- **Don't** expose `_harmonic_created_by` or `_harmonic_created_at` in `query_rows` markdown output. Only `_harmonic_row_id` is needed for discoverability; the other two stay internal.
- **Don't** edit `CHANGELOG.md` in feature-branch commits — handle post-merge.

danallison:
from review agent:

# Round-2 review fixes for the `_harmonic_` rename

Address five findings from review of commits `ac788d2d`, `e9dcd5c8`, `7a3f6971` on branch `melody/table-row-id-discoverable`. Read each finding and decide the smallest fix; don't expand scope.

## 1. Migration silently swallows per-note failures

`db/migrate/20260628000000_rename_table_row_internal_fields_to_harmonic_prefix.rb`

Currently a per-note `rescue StandardError` logs at `warn` with no summary, so partial failure looks like success.

- Count failures into a local counter.
- Log at `Rails.logger.error` (not `warn`).
- After `find_each` completes, if any failures occurred: `raise "..."` with the count and a sample of skipped note ids, so the deploy fails loud rather than continuing with inconsistent data.

## 2. Inconsistent exception type for the `_harmonic_` violation

Same condition raises different classes in different layers:

- `app/services/api_helper.rb:339` → `ArgumentError`
- `app/services/note_table_service.rb:48` → bare `RuntimeError`

Standardize on `ArgumentError` (caller error, not runtime failure) in both places. Then:

- Update `app/controllers/notes_controller.rb` rescue clauses for `add_column` and `batch_table_update` to include `ArgumentError` alongside the existing `RuntimeError, ActiveRecord::RecordInvalid`. The `create_table_note` rescue at line 133 already has it.

## 3. `define_columns!` lacks the prefix guard

`app/services/note_table_service.rb:37`

`add_column!` short-circuits on `_harmonic_`; `define_columns!` does not, so the same input raises `RuntimeError` via one path and `ActiveRecord::RecordInvalid` via the other.

Pick **one** of:

- **(Preferred)** Drop the explicit guard from `add_column!` and rely on the model validator at `app/services/note_table_validator.rb:63` as the single source of truth. Update `add_column!`'s test in `test/services/note_table_service_test.rb` to expect `ActiveRecord::RecordInvalid`.
- Add a matching `start_with?("_harmonic_")` guard to `define_columns!` that raises `ArgumentError`. Add a test mirroring the `add_column!` one.

## 4. Redundant transaction wrapping a single UPDATE

`db/migrate/20260628000000_rename_table_row_internal_fields_to_harmonic_prefix.rb:53-55`

Drop the `ActiveRecord::Base.transaction { ... }` wrapper. `update_columns` is a single statement and already atomic.

## 5. Reject-test assertion is loose

`test/controllers/notes_controller_test.rb`, the new test `create_table_note action rejects a column using the reserved _harmonic_ prefix` (around line 547).

Add `assert_response :unprocessable_entity` (or whatever status `render_action_error` returns for this path — check one of the existing rejection tests in the file for the convention) so the test pins the failure mode, not just the message text.

PR #285, opened 2026-06-28
Notification settings UI for humans and agents (#265)

### LINKED ISSUE #265: notifications settings for both humans and agents
backend wiring is mostly there, but currently no way for users to change their notification settings

melody-harmonic:
Closes #265.

## What

The backend for per-type, per-channel notification preferences already existed on `TenantUser` (`settings.notification_preferences`, read by `NotificationDispatcher.channels_for_user`), but there was no user-facing way to change it. This adds that surface for **both human users and AI agents**.

## Changes

- **`TenantUser`** — `NOTIFICATION_TYPE_LABELS` / `NOTIFICATION_CHANNELS` constants and `update_notification_preferences!(prefs)`: a **merge-not-replace** bulk update that ignores unknown types/channels.
- **`NotificationPreferencesParams` concern** — parses the payload. The HTML form submits a *complete* matrix (browsers omit unchecked boxes, so those are recorded as `false`); the markdown action submits a *partial* object and only the supplied keys change.
- **`UsersController` / `AiAgentsController`** — an HTML update action plus markdown `describe_`/`execute_` actions, wired into `ActionsHelper` (`ACTION_DEFINITIONS` + both `/settings` route entries) and `config/routes.rb`.
- **Views** — a shared `_notification_preferences` checkbox matrix on both HTML settings pages, plus a current-state table and action pointer in both `.md` settings views.

## Markdown action

`update_notification_preferences` takes a `notifications` object, e.g.:

```json
{ "comment": { "email": true }, "mention": { "email": false } }
```

Only the keys you include are changed (partial merge), so an agent can flip a single channel without restating the whole matrix.

## Tests

Model merge semantics (multi-type update, merge preserves untouched types, unknown keys ignored) and the HTML + markdown paths for both users and agents, including the cross-user authorization guard.

## Caveat

⚠️ **Unverified by execution** — no Docker/Ruby in my environment, so I couldn't run the suite or `srb tc`/`rubocop`. Written to existing patterns (mirrors `update_profile`); the repo's bash static-analysis checks (debug-code, secrets, tenant-safety, style-guide) report no new violations. Worth a green CI run before merge.

## Follow-ups (left out)

- The `email` channel assumes the account has an email address; for agents that's often unset. Could add a hint/disable when no email is on file.
- No per-collective overrides — preferences are per-`TenantUser` (per tenant), matching the existing backend.

### DIFF
--- a/app/services/actions_helper.rb
+++ b/app/services/actions_helper.rb
@@ -569,6 +569,18 @@ class ActionsHelper
       authorization: [:self, :representative],
       visibility: :public,
     },
+    "update_notification_preferences" => {
+      description: "Update which channels deliver each notification type. Pass a `notifications` object mapping type to channel booleans; only the keys you include are changed.",
+      params_string: "(notifications)",
+      params: [
+        { name: "notifications", type: "object",
+          description: "Nested object { \"<type>\": { \"in_app\": true|false, \"email\": true|false } }. " \
+                       "Types: mention, comment, participation, system, reminder, chat_message, trio_unavailable, tune_in, trustee_authorization. " \
+                       "Channels: in_app, email. Example: { \"comment\": { \"email\": true }, \"mention\": { \"email\": false } }.", },
+      ],
+      authorization: [:self, :representative],
+      visibility: :public,
+    },
     "update_scratchpad" => {
       description: "Update your scratchpad with notes for your future self",
       params_string: "(content)",
@@ -1419,6 +1431,8 @@ class ActionsHelper
       actions: [
         { name: "update_profile", params_string: ACTION_DEFINITIONS["update_profile"][:params_string],
           description: ACTION_DEFINITIONS["update_profile"][:description], },
+        { name: "update_notification_preferences", params_string: ACTION_DEFINITIONS["update_notification_preferences"][:params_string],
+          description: ACTION_DEFINITIONS["update_notification_preferences"][:description], },
       ],
     },
     "/u/:handle/lists" => {
@@ -1474,6 +1488,8 @@ class ActionsHelper
       actions: [
         { name: "update_profile", params_string: ACTION_DEFINITIONS["update_profile"][:params_string],
           description: ACTION_DEFINITIONS["update_profile"][:description], },
+        { name: "update_notification_preferences", params_string: ACTION_DEFINITIONS["update_notification_preferences"][:params_string],
+          description: ACTION_DEFINITIONS["update_notification_preferences"][:description], },
       ],
     },
     "/ai-agents/:handle/bridge-setup" => {
--- a/app/services/actions_helper.rb
+++ b/app/services/actions_helper.rb
@@ -570,7 +570,9 @@ class ActionsHelper
       visibility: :public,
     },
     "update_notification_preferences" => {
-      description: "Update which channels deliver each notification type. Pass a `notifications` object mapping type to channel booleans; only the keys you include are changed.",
+      description: "Update which channels deliver each notification type. " \
+                   "Pass a `notifications` object mapping type to channel booleans; " \
+                   "only the keys you include are changed.",
       params_string: "(notifications)",
       params: [
         { name: "notifications", type: "object",
--- a/app/services/actions_helper.rb
+++ b/app/services/actions_helper.rb
@@ -581,7 +581,7 @@ class ActionsHelper
                        "Channels: in_app, email. Example: { \"comment\": { \"email\": true }, \"mention\": { \"email\": false } }.", },
       ],
       authorization: [:self, :representative],
-      visibility: :public,
+      visibility: :private,
     },
     "update_scratchpad" => {
       description: "Update your scratchpad with notes for your future self",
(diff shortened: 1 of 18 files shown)


danallison:
currently agents do not have real email addresses, so it doesn't make sense to have any email notifications enabled for an agent. let's update the ui for agent settings to be a simple boolean toggle. either the notification is on or it's off. only human users need to distinguish between in-app and email.

danallison:
in agent settings, there should not be a separate "save notification preferences" button. there should only be one "save settings" button for the whole settings page, which includes notification preferences.

danallison:
it looks like the agent notifications ui is only on the agent settings page and not the new agent form page. it should be on both.

danallison:
from the review agent

# Notification settings UI — review fixes

Branch: `melody/notification-settings-ui`. Fix in red-green order. Test only touched files locally; CI runs the suite.

## Correctness

### 1. Partial write in `update_settings` (high priority)
`app/controllers/ai_agents_controller.rb:180` — `apply_agent_notification_preferences_from_form` calls `tenant_user.save!` BEFORE `@ai_agent.save` at line 182. If the agent save fails, prefs are already committed and the user sees "error." Make the notification write conditional on agent save success, or wrap both in one transaction. Test: invalid `@ai_agent` change + notification toggles → tenant_user prefs unchanged.

### 2. Unchecked boxes stored as JSON null, not `false`
`app/controllers/concerns/notification_preferences_params.rb:31` — `ActiveModel::Type::Boolean.cast(nil)` returns `nil`. When `complete: true` and a channel is absent from params, this writes nil into a column typed `T::Hash[String, T::Boolean]`. `notification_enabled?` masks it via `== true`, but the contract is violated. Coerce the absent-channel path to `false`. Test: submit HTML form with one box unchecked → stored value is `false` (use `==` against `false`, not `!`, to catch nil).

## Cleanup / altitude

### 3. Three sources of truth for notification types
`app/models/tenant_user.rb` — `Notification::NOTIFICATION_TYPES`, `DEFAULT_NOTIFICATION_PREFERENCES.keys`, and the new `NOTIFICATION_TYPE_LABELS.keys` all enumerate the same 9 keys; the new comment admits the drift risk. Either derive one set from another, or add a test asserting all three are equal.

### 4. Dead standalone HTML endpoint
`app/controllers/ai_agents_controller.rb:592` (`update_notification_preferences`) + the route at `config/routes.rb`. No view links to it — the agent UI folds notifications into the main settings form. Delete the action, route, and its test, unless you want a second submit path (you don't — the UX is folded). The markdown action surface stays.

### 5. `simple:` mode duplicates header + body in the fields partial
`app/views/shared/_notification_preferences_fields.html.erb` — branches at lines 24 and 37 with near-identical cell markup; aria-label format already drifts between the two. Normalize to a channels list at the top (`simple ? %w[in_app] : NOTIFICATION_CHANNELS`) and iterate once.

### 6. Pick one stance on `tenant_user` nil
`app/controllers/ai_agents_controller.rb` lines 594, 613, 645 explicitly guard `tu.nil?`, but `app/views/ai_agents/settings.md.erb:100` and the partial caller in `settings.html.erb` dereference `@ai_agent.tenant_user` without a guard. If nil is reachable, the views crash; if it isn't, the controller checks are dead code. Pick one and apply it consistently.

## Conventions (CLAUDE.md)

### 7. Line >150 chars
`app/services/actions_helper.rb:573` — 178 chars. Break the `description:` string across lines like the `params` description below it.

### 8. Stale comment on create seeding
`app/controllers/ai_agents_controller.rb:463-464` — comment claims "HTML only — the marker is absent on markdown/API creation." The code doesn't gate by format; any caller passing `notifications_present` triggers seeding. Rewrite or delete.

## Lower priority (judgment calls — flag, not require)

- **Human form has no `notifications_present` marker** (`app/views/shared/_notification_preferences.html.erb`). The agent form guards against empty submits; the human form doesn't. Exploitable only via odd client behavior, so it's defense-in-depth, not a current bug. Mirror the agent guard if you want symmetry.
- **Test brittleness in "create applies notification toggles"** (`test/controllers/ai_agents_controller_test.rb:443`). Lookup is `User.where(user_type: "ai_agent").order(:created_at).last` — works today inside the transactional test, but scope-leaky. Scope to `@user.ai_agents` or look up by the submitted name for clarity.

app/services/actions_helper.rb
```diff
@@ -569,6 +569,20 @@ class ActionsHelper
...
+                   "only the keys you include are changed.",
+      params_string: "(notifications)",
+      params: [
+        { name: "notifications", type: "object",
+          description: "Nested object { \"<type>\": { \"in_app\": true|false, \"email\": true|false } }. " \
+                       "Types: mention, comment, participation, system, reminder, chat_message, trio_unavailable, tune_in, trustee_authorization. " \
+                       "Channels: in_app, email. Example: { \"comment\": { \"email\": true }, \"mention\": { \"email\": false } }.", },
+      ],
+      authorization: [:self, :representative],
+      visibility: :public,
```
danallison: visibility should be `private`. these preferences are only visible to the user.

melody-harmonic: Done in `33f9f34e` — changed to `visibility: :private` and moved the `EXPECTED_VISIBILITY` lock-in entry to the private group (plus the notification-actions spot-check). Authorization and the CapabilityCheck block are unchanged.

PR #289, opened 2026-06-29
Case-insensitive, case-preserving handles (citext) — handle-unification Goal 1

melody-harmonic:
Implements **Goal 1** of `.claude/plans/handle-model-unification.md`: GitHub-style handles — the display form remembers the case the user chose, while lookup and uniqueness are case-insensitive.

`@linus`, `@LINUS`, `@Linus` all resolve to the same identity; `Linus` can't be registered if `linus` exists; `/u/Linus` and `/u/linus` route to the same person.

## How

Convert `tenant_users.handle` and `collectives.handle` from `varchar` to the Postgres **`citext`** type. citext compares and indexes case-insensitively, which means:

- the existing `UNIQUE (tenant_id, handle)` indexes now reject a cased duplicate **with no index change** (Postgres rebuilds them with the citext comparator);
- all ~40 `find_by(handle:)` / `where(handle:)` callsites become case-insensitive **with no Ruby change**;
- the stored value keeps its original case, so profiles/mentions render `Linus`.

### Model changes
- `TenantUser#normalizes` → `parameterize(preserve_case: true)`: `"Jane Smith"` → `"Jane-Smith"` (was `"jane-smith"`). Auto-generated *default* handles stay lowercase (unchanged behavior).
- `Collective` handle validation accepts uppercase (`[a-zA-Z0-9-]`).
- Reserved-handle checks (`trio`, `main`) are case-folded so `"Trio"`/`"Main"` can't slip past them.

### Schema
Migration `20260702000000_make_handles_case_insensitive` enables the `citext` extension and retypes both `handle` columns to `public.citext`.

CI prepares the test DB with `db:schema:load` (loads `db/structure.sql`), **not** by running migrations, so `db/structure.sql` has to reflect the change. Rather than hand-editing the dump, it's **regenerated by actually running the migration** — load `main`'s schema, `db:migrate`, commit the resulting dump — so what's committed is machine-generated and there's no doubt the migration produces it. The migration is timestamped **after `main`'s latest (`20260701000000`)** so it applies in order on a DB that's already migrated past that point. Generated Sorbet RBIs are unchanged (citext maps to `::String`).

### Data safety
Handles were lowercased on save until now, so no two existing rows differ only by case — the type change can't violate the unique indexes.

## Tests
Added model coverage (case preserved on explicit set; case-insensitive uniqueness; case-insensitive lookup; cased reserved handles rejected) for both `TenantUser` and `Collective`, and updated the two existing assertions whose mixed-case inputs now preserve case (`signup` invite-accept, TenantUser rename).

## Verification
Ran locally against the dev Postgres stack: the migration applies cleanly, and the affected suites — `tenant_user_test`, `collective_test`, `signup_controller_test` — are green (190 runs, 472 assertions, 0 failures). CI is green across static-analysis/Sorbet, TypeScript, and all four rails-test groups.

## Goal 2 — stacked follow-up (#290)
The plan's second goal — unify the user/collective handle namespaces so a collective's `identity_user` shares the collective handle (today it's `SecureRandom.hex(16)`; `User#handle` only *fakes* `collectives/<handle>` at the display layer) — is now open as **#290**, stacked on this branch. It reuses this citext foundation and backfills existing identity users to their collective's handle, resolving the rare tenant-local collision with a numeric/hex suffix. Merge this first, then #290 retargets to `main`.


danallison:
are there going to be any issues with this migration if the schema change list is out of order? also are we sure that cloud-managed postgres allows this extension?

PR #290, opened 2026-06-29
Unify collective + identity-user handle namespace — handle-unification Goal 2

melody-harmonic:
Implements **Goal 2** of `.claude/plans/handle-model-unification.md`. Builds on **#289** (Goal 1, citext), now **merged** — this branch is rebased directly on `main`.

## What changes
Today a collective `foo-team` has an `identity_user` (a `collective_identity` `TenantUser`) minted with a **random hex handle** — two different handles for what users think of as one entity. This makes them share one handle.

- **`TenantUser.identity_handle_for`** — picks a tenant-unique handle from the collective's own handle, case-preserving, falling back to a `-XX` numeric suffix on collision (existing user already holds it, or a reserved handle the identity can't claim).
- **`create_identity_user!`** seeds the identity from the collective handle instead of `SecureRandom.hex(16)`. (Also: returns early on a blank handle rather than minting an orphan identity — the collective fails `handle_is_valid` anyway.)
- **`after_update` sync** keeps the identity handle in lockstep when a collective is renamed.
- **`Collective.handle_available?`** (the `GET /collectives/available` endpoint behind the New Collective form) now checks the **unified namespace** — both `collectives` *and* `tenant_users` — instead of collectives only. See below.
- **Backfill migration** (`20260702010000`) converts existing hex identity handles to the collective handle, raw-SQL so it's independent of model code/tenant scoping. Data-only — no schema change — so its only effect on `db/structure.sql` is the `schema_migrations` entry.

## Handle-availability endpoint
Because collective creation now seeds an identity user from the collective's handle, the availability check has to see the whole namespace. Previously `handle_available?` only asked `Collective.where(handle:)`, so the form reported `foo` as **available** even when a *user* `foo` existed — the collective would then be created with a suffixed identity (`foo-team-XX`), reintroducing exactly the collective≠identity mismatch Goal 2 removes. The check now also consults `tenant_users`, so an "available" handle guarantees the new collective's identity user can take the *identical* handle. Both columns are citext, so the check stays case-insensitive. Suffixing remains the safety net for the API path and backfill; the form just steers humans toward clean handles.

## Collision policy
Per the tracking vote (decision `72345628`, deadline 2026-07-08): **suffix** (my starred pick). Backfill collisions (a human already holds the collective's handle, from before the namespaces were unified) and any going-forward collision both resolve to `foo-team-XX`.

## Cross-namespace uniqueness (plan bullet 3)
Every collective's identity user occupies the collective's handle in the shared `(tenant_id, handle)` `tenant_users` unique index, so a human can no longer register a handle a collective holds. The residual edge — a *new* collective whose handle a human already holds — resolves by suffixing the identity (the human keeps `@handle`), consistent with the suffix policy; the availability endpoint (above) now reports that handle as taken so the form nudges the creator to a clean one.

## Schema
Like #289, `db/structure.sql` is **regenerated by running the migration** (load schema, `db:migrate`, commit the dump), not hand-edited. The migration is timestamped after #289's (`20260702000000`) and `main`'s latest, so the stack applies in order. Post-rebase both `20260702000000` and `20260702010000` are present in `schema_migrations` and the citext extension is in the dump.

## Tests
`test/models/collective_test.rb`: identity shares the handle (case-preserved); suffixing on user collision and on a reserved handle; rename-sync; and `handle_available?` returns false when a *user* already holds the handle (incl. a case-insensitive variant). Plus the existing collective/tenant_user handle suites.

## Verification
Ran locally against the dev Postgres stack: `collective_test` + `tenant_user_test` green (**156 runs, 336 assertions, 0 failures**), `collectives_controller_test` green (111 runs, 367 assertions, 0 failures), and `srb tc` clean. Now that the base is `main`, GitHub CI is running on this PR.

### DIFF
--- a/app/models/collective.rb
+++ b/app/models/collective.rb
@@ -103,9 +103,15 @@ class Collective < ApplicationRecord
   def self.handle_available?(handle)
     return false if RESERVED_HANDLES.include?(handle.to_s.downcase)
 
-    # `handle` is citext, so this count is case-insensitive: "Foo" is taken
+    # Collective and user handles share one per-tenant namespace (Goal 2 of
+    # handle-model-unification): creating a collective also seeds an identity
+    # user that claims the collective's handle. Treat a handle as available only
+    # when neither a collective nor a user already holds it, so the new
+    # collective's identity user can take the identical handle instead of a
+    # suffixed fallback (@foo-team ↔ /collectives/foo-team stay in sync). Both
+    # columns are citext, so the lookups are case-insensitive: "Foo" is taken
     # once "foo" exists.
-    Collective.where(handle: handle).count == 0
+    !Collective.where(handle: handle).exists? && !TenantUser.where(handle: handle).exists?
   end
 
   sig { void }
--- a/app/models/collective.rb
+++ b/app/models/collective.rb
@@ -14,6 +14,7 @@ class Collective < ApplicationRecord
   belongs_to :trio_user, class_name: "User", optional: true
   before_validation :create_identity_user!
   before_create :set_defaults
+  after_update :sync_identity_user_handle!, if: :saved_change_to_handle?
   tables = ActiveRecord::Base.connection.tables - [
     "tenants", "users", "tenant_users",
     "collectives", "api_tokens", "oauth_identities",
@@ -643,6 +644,10 @@ class Collective < ApplicationRecord
     return if private_workspace?
     return if chat?
     return if identity_user
+    # The identity shares the collective's handle; without one there's nothing
+    # to share, so defer to `handle_is_valid` to surface the blank-handle error
+    # rather than minting an orphan identity user with a placeholder handle.
+    return if handle.blank?
 
     identity = User.create!(
       name: name,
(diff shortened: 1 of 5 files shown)


melody-harmonic:
Addressed the review's minor note in eb8fe6b4.

The backfill migration only suffixed identity-user handles on collision, whereas the runtime `TenantUser.identity_handle_for` also suffixes when the desired handle is reserved. That gap is actually reachable (not just defensive): collective validation forbids only `Collective::RESERVED_HANDLES` (`["main"]`), **not** the tenant_users reserved set (`TenantUser::RESERVED_HANDLES`, `{"trio"=>"trio"}`). So a collective named `trio` is permitted, and its identity-user backfill would have claimed the reserved `trio` handle that the live path suffixes away — the collision check only catches it in tenants that already hold a live `trio` tenant_user.

Fix derives `desired` exactly as the runtime helper does — `parameterize(preserve_case: true)` + reserved-suffix via the shared `RESERVED_HANDLES` constant — so the backfill now produces precisely what a fresh `create_identity_user!` would. Unverified by local execution (no Docker here); relying on CI.

danallison (approved):

Issue #291, opened 2026-06-29
@ mentions should render as links


danallison:
@ mention handles should link to the mentioned user's profile when rendered

PR #295, opened 2026-06-29
Render @mentions as profile links

melody-harmonic:
## What

Implements #291 — `@handle` mentions now render as links to the mentioned user's profile.

When markdown content is rendered to HTML, any `@handle` that resolves to a real user in the current tenant is rewritten as `<a href="/u/handle" class="mention-link">@handle</a>`. Handles that don't match a user are left as plain text.

## How

- **`MentionParser.resolve_paths(text, tenant_id:, collective:)`** — returns a `{ handle => profile_path }` map for handles that resolve to a user, mirroring `.parse`'s resolution logic, including the collective-local `@trio` special case (so `@trio` links to `/u/trio`).
- **`MarkdownRenderer#linkify_mentions`** — runs as the final render step. It walks the rendered HTML's text nodes, **skipping** `<a>`, `<code>`, and `<pre>` so mentions inside existing links or code stay untouched. Literal text is HTML-escaped before the mention pattern is applied (handle chars are never HTML-special, so the pattern still matches). Cheap early-outs when there's no tenant context or no `@` in the content.
- Adds a `.mention-link` style to `github-markdown-overrides.css`.

Applies to both `render` and `render_inline`.

## Tests

- `MentionParser.resolve_paths`: blank input, resolvable handle → path, unknown handle omitted, `@trio` → `/u/trio`.
- `MarkdownRenderer`: links resolvable mentions, leaves unknown mentions plain, multiple mentions, no linking inside inline code / code blocks / existing links, and correct HTML escaping of surrounding text.

## Verification

I can't run the suite from this VM (Docker-based) — relying on CI. Did static checks: traced `user.path`/`tenant_user.path` (`/u/<handle>`), confirmed `Tenant.current_id`/`Collective.current_id` thread-state APIs, and structured the gsub to avoid Sorbet nilable-match issues.


danallison:
this seems overengineered. why not just preprocess the markdown to replace `@handle` with `[@handle](/u/handle)` and let the markdown renderer render it like any other link?

melody-harmonic:
The Nokogiri walk isn't about resolution (preprocess could resolve too) — it's about *not* linkifying `@handle` in contexts where it shouldn't become a link. A raw-markdown regex is blind to markdown structure, so a `@handle → [@handle](/u/handle)` substitution would fire inside:

- **code spans / fenced blocks** — `` `@danallison` `` or a code sample documenting mention syntax would turn into a link inside `<code>`;
- **existing links** — `[ping @danallison](/x)` becomes a nested `[ping [@danallison](/u/danallison)](/x)`, which markdown renders broken;
- **autolinked emails** — `MENTION_PATTERN` has no left boundary, so `daniel@example.com` matches `@example`. With `autolink: true` those are already `<a mailto:>` so the walk skips them; a preprocess would rewrite mid-email.

That said, you're right the parse + ancestor walk is heavier than needed. The middle ground that keeps correctness without it: subclass the Redcarpet renderer and override `normal_text(text)` to do the substitution there. Redcarpet only calls `normal_text` for plain text runs — code spans (`codespan`) and blocks (`block_code`) bypass it, and autolinked emails go through the separate `autolink` callback — so both problem cases above disappear with no HTML re-parsing. The one case it *doesn't* cover is a mention inside existing link text (you'd get a nested anchor), which is rare and easy to guard.

Want me to refactor to the `normal_text` approach? Drops the extra Nokogiri pass and the `MENTION_SKIP_ANCESTORS` logic, which I think addresses the overengineering concern while keeping the code/link/email behavior.

danallison:
yes

melody-harmonic:
Done — refactored to the `normal_text` approach in d7ed31f.

- `MentionRenderer < Redcarpet::Render::HTML` linkifies handles in `normal_text`. Code spans, fenced blocks, and autolinked emails go through other Redcarpet callbacks, so they're left literal for free — no more `MENTION_SKIP_ANCESTORS` and no post-render Nokogiri pass.
- The only context `normal_text` can't see is link text, so `link` unwraps any mention anchor in its content before building the enclosing link (keeps the `[@alice](url)` → no-nested-anchor behavior). It calls `super`, so `safe_links_only` etc. stay intact.
- Mentions are now emitted during the markdown pass (before `sanitize`), so like every other relative link they pick up `rel="noopener noreferrer"`. Updated the three exact-match mention tests to tolerate that.
- Added `sorbet/rbi/shims/redcarpet.rbi` declaring the two C-defined callbacks so the overrides + `super` typecheck under `srb tc`.

One tradeoff worth flagging: resolution now happens per plain-text run (one `resolve_paths` query per run that contains `@`) instead of one batched query for the whole document. The `text.include?("@")` early-out skips the vast majority of runs, so for typical notes/comments it's 1–2 queries — but a document that mentions handles across many separate paragraphs would issue more queries than the old single-pass did. If you expect mention-heavy content, I can add a per-render memo (would mean instantiating the renderer per render rather than sharing it); left it out to keep the diff minimal. Let me know.

Still no Docker on this VM so I can't run the suite locally — leaning on CI, will follow up once it's green.

app/assets/stylesheets/github-markdown-overrides.css
```diff
@@ -18,4 +18,16 @@
...
+/* @mention links. MarkdownRenderer rewrites @handle mentions that resolve to
+   a real user into profile links with a .mention-link class. Render them with
+   a slightly heavier weight so they read as a mention, not an ordinary link. */
+.markdown-body a.mention-link {
+  font-weight: 600;
+  text-decoration: none;
+}
+
+.markdown-body a.mention-link:hover {
+  text-decoration: underline;
```
danallison: why do we need special style for mention links? it's just a link, right?

danallison:
what happens if someone tags a non-existent user? does that still get rendered as a link, or does it correctly render as plain text?

melody-harmonic:
Plain text — correctly. In `normal_text`, resolution goes through `MentionParser.resolve_paths`, which returns a hash containing **only** handles that map to a real user in the tenant; handles that don't resolve are omitted. The substitution then does:

```ruby
path = paths[handle]
path ? %(<a href="..." class="mention-link">@#{handle}</a>) : match
```

so an unresolved handle falls to the `: match` branch and the original `@handle` text is emitted verbatim — no anchor, no `.mention-link` class. (And because `escaped` is HTML-escaped before the gsub, it stays escaped plain text, not raw.)

Covered by `test "render leaves unknown @mentions as plain text"`:

```ruby
html = MarkdownRenderer.render("Hello @nobody_here", display_references: false)
assert_includes html, "@nobody_here"
assert_not html.include?(%q{class="mention-link"})
```

Still leaning on CI to actually run it (no Docker locally).

melody-harmonic:
CI is green ✅ (services / controllers / integration / models / static-analysis / typescript all pass).

Heads-up on a bug my first push of this refactor had, now fixed in d62e428: I'd written `#link` to call `super`, but Redcarpet's render callbacks are C methods with no Ruby superclass method, so it raised `NoMethodError` at runtime on every rendered link. `srb tc` passed because I'd added a shim declaring the callbacks — a good reminder that the Sorbet shim asserts a method exists without proving it's `super`-able. Fixed by building the anchor directly in `#link` (href protocol safety is still enforced in `.sanitize`, same as before) and dropping the shim.

danallison:
see https://github.com/ibis-coordination/Harmonic/pull/295#discussion_r3494979786
please remove unnecessary css

app/assets/stylesheets/github-markdown-overrides.css
melody-harmonic: Right — removed it (db72fb9). Mentions now render as ordinary links, no special styling.

I kept the `.mention-link` class on the anchor though, because it turned out to be load-bearing: the refactored renderer's `MENTION_LINK` regex matches `class="mention-link"` to detect and unwrap a mention anchor that lands inside link text (the `[@alice](url)` no-nested-anchor case). It's also what the renderer tests assert on to prove a mention got linkified vs. it just being an ordinary markdown link. It carries no CSS now — purely a marker. Happy to drop it too if you'd rather mentions be truly indistinguishable from any other link, but that'd mean reworking the unwrap logic and loosening those tests.

PR #303, opened 2026-06-30
Fix #287: redirect summary pages to a canonical route so confirm-read works

### LINKED ISSUE #287: Cannot confirm reading summary notes — confirm-read POSTs to nonexistent <parent>/summary/confirm.html route
## Summary

You cannot confirm reading a **summary note** from its summary page (`<parent>/summary`). The "Confirm Read" button POSTs to a route that doesn't exist, so the confirmation silently fails. As @danallison noticed, this stems from the special "unique summary path" that summary notes use.

## Root cause

`Note#path` is overridden so that a summary note returns its parent's canonical summary URL instead of its own `/n/:id` URL:

```ruby
# app/models/note.rb:241
def path
  if is_summary? && summarizable
    "#{summarizable.path}/summary"   # e.g. /collectives/foo/n/<parent_id>/summary
  else
    super                             # canonical /n/<id>
  end
end
```

The confirm-read UI builds its POST target from `@note.path`:

```erb
<%# app/views/notes/_confirm.html.erb %>
data-url="<%= @note.path + '/confirm.html' %>"
```

On a summary page, `render_summary_for` sets `@note` to the summary note (`application_controller.rb:1123`), so the button posts to:

```
/collectives/foo/n/<parent_id>/summary/confirm.html
```

But the only confirm/acknowledge routes are nested under the note resource (`/n/:id/...`), **not** under `/summary` (`config/routes.rb:589-637`):

```ruby
resources :notes, only: [:show], path: "#{prefix}/n" do
  get  '/summary'        => 'notes#summary'   # no nested action routes
  post '/confirm.html'   => 'notes#confirm_and_return_partial'
  post '/actions/confirm_read' => 'notes#confirm_read'
  ...
end
```

There is no `/n/:id/summary/confirm.html` route, so the POST 404s and the read is never recorded.

The same override breaks the agent/markdown path too: the action frontmatter on `<parent>/summary` advertises `confirm_read` (and `acknowledge`, `report`, etc.) at `<parent>/summary/actions/...`, which likewise don't resolve.

## Reproduction

1. Add a summary to any note/decision/commitment (`add_summary`).
2. Navigate to `<parent>/summary`.
3. Click **Confirm Read**.
4. The request fails (no matching route); the confirmation is not recorded.

## Suggested fix

The codebase already distinguishes the canonical resource URL (`path`, used to build action endpoints) from the display URL (`display_path`, used for links) — see `application_record.rb:168-174`, where comments override `display_path` and leave `path` canonical.

Summary notes do the opposite: they override `path` itself, which is what breaks action routing. Preferred direction:

- **Move the `<parent>/summary` override from `Note#path` to `Note#display_path`**, leaving `path` canonical (`/n/:id`). Then confirm/acknowledge/report forms and the markdown action list resolve correctly, while breadcrumbs/links can keep using `display_path`. Audit callers that currently rely on `path` returning the summary URL (e.g. agents fetching `<parent>/summary`) and point them at `display_path`.

Alternatives: (a) build the action URLs in the confirm/acknowledge partials and md action list from the note's canonical `/n/:id` path instead of `@note.path`; or (b) add nested action routes under `/summary` (more surface area, duplicative).

## Affected files

- `app/models/note.rb` (`#path` override, ~L241)
- `app/views/notes/_confirm.html.erb` (`confirm.html` / `acknowledge.html` targets)
- `config/routes.rb` (~L589-637)
- `app/controllers/application_controller.rb` (`render_summary_for`, ~L1115)

_Filed by melody on behalf of @danallison. Not yet verified against the test suite._

melody-harmonic:
Fixes the primary repro in #287: the **Confirm Read** button on a summary page silently 404'd, so the read was never recorded — and it also covers the agent-dispatch half of the issue.

## Root cause
`Note#path` was overridden to return `<parent>/summary` for summary notes. But the confirm/acknowledge/report partials build their POST target by suffixing onto `@note.path` (e.g. `@note.path + '/confirm.html'`). On a summary page `@note` is the summary note, so the button posted to `<parent>/summary/confirm.html` — a route that doesn't exist (action routes are nested under `/n/:id`, not `/summary`). Agents hit the same wall: the MCP layer suffixes `/actions/<name>` onto whatever URL the agent fetched, so an agent sitting on `<parent>/summary` dispatched to a non-existent route too.

## Fix
Per @danallison's review: rather than juggling which of `path`/`display_path` carries the `<parent>/summary` divergence, **drop the special summary path entirely.** A summary is an ordinary note that lives at its canonical `/n/:id` page; `<parent>/summary` becomes a redirect-only entry point.

- `ApplicationController#render_summary_for` now **302-redirects** `<parent>/summary` to the summary note's canonical `/n/:id` page, preserving the requested format (`<parent>/summary.md` still lands on markdown). When no summary exists yet it still renders the `shared/no_summary` page (nothing to redirect to).
- `Note#path` **and** `Note#display_path` both return the canonical `/n/:id` for summaries — the `<parent>/summary` override is gone from the model. Every suffix-built action endpoint (confirm / acknowledge / report, the markdown action list) now resolves.
- This is what fixes the confirm-read 404, and it closes the **agent dispatch path** too: an agent that fetches `<parent>/summary` follows the redirect to the canonical page, whose action list already points at `/n/:id/actions/…` — no `/summary`-nested action routes needed.
- `_summaries_section` links the summary at its canonical path.
- Removed the now-orphaned `shared/summary.{html,md}.erb` templates (the canonical note show page already renders `shared/_note_main`) and updated the partial doc comments.

You can still navigate straight to `<parent>/summary` — the URL just bounces you to the canonical page.

## Tests
Red-green by construction. **Unverified locally — this environment has no Docker, so the suite/rubocop/sorbet run only in CI.** (CI did real work here: it caught two bugs in the regression test itself — a hardcoded path that broke on the main collective whose `Collective#path` is nil, and an `assert_select` arg I'd passed as a message when it's a text-matcher. Production code was correct; both were test-only fixes. CI is now fully green.)
- `note_test`: summary `#path` and `#display_path` are both the canonical `/n/:id`.
- `add_summary_action_test`: `<parent>/summary[.md]` redirects to the summary's canonical page and following the redirect serves the content; the Confirm Read button on the landed page targets `/n/:id/confirm.html`, and POSTing there records the read end-to-end.


danallison:
all display links should use the display path, especially in markdown views. the whole point of having a special `/summary` path is so agents can discover the pattern and go straight to an item's summary without having to know the summary's id.

danallison:
actually i've changed my mind. this special path is more trouble than it's worth. let's just use the canonical path with the id and have the `/summary` paths redirect. that way it's still possible to navigate straight to a summary without knowing the id, but we don't have to deal with the headache of all of these special case action routes.

melody-harmonic:
Done — switched #303 to the redirect approach in 12a723c6.

**What changed vs. the first cut:**
- `<parent>/summary` (and `.md`) now **302-redirects** to the summary note's canonical `/n/<id>` page (format preserved). No summary yet → still renders the no-summary page, since there's nothing to redirect to.
- Dropped the special path from the model entirely: both `Note#path` **and** `Note#display_path` return the canonical `/n/<id>` for summaries. No more juggling which accessor carries the `/summary` divergence.
- Removed the now-orphaned `shared/summary.{html,md}.erb` templates — the canonical note show page already renders `shared/_note_main`.

**This also closes the agent half I'd flagged.** An agent that fetches `<parent>/summary` follows the redirect to the canonical page, whose action list already points at `/n/<id>/actions/…` — so confirm-read resolves for agents with **no** `/summary`-nested action routes. That's the headache you wanted gone.

You can still navigate straight to a summary without knowing its id; the URL just bounces you to the canonical page.

Same caveat as always: tests are red-green by construction but **unverified locally** (no Docker here) — leaning on CI. Watching it now.

danallison:
please update the pr description to reflect the final state

melody-harmonic:
Done — rewrote the PR title and description to match the final redirect-based approach. The body now leads with the redirect entry point (`<parent>/summary` 302s to the canonical `/n/:id`, format preserved), notes that both `Note#path` and `display_path` return the canonical URL so the `/summary` override is gone from the model, and records that this closes the agent-dispatch half too (no `/summary`-nested action routes needed). The stale "move the divergence to display_path" framing and the open follow-up caveat are gone, since the redirect covers both paths.

PR #304, opened 2026-06-30
Fix #267: block final statements until a decision is fully resolved

### LINKED ISSUE #267: Disallow final statements until decision is fully resolved (not just past deadline)
## Problem

A decision's final statement can currently be written as soon as the **deadline** passes, even if the result is not yet **fully resolved**. For vote/lottery decisions whose top options are tied, the winner is not known until the drand tiebreaker **beacon** is drawn — and that draw happens *asynchronously, after* the deadline. This opens a window where a final statement is written against a provisional (unresolved) ordering that can still flip.

This actually bit us: a tie between two options had a provisional sort showing option A on top, a final statement was written closing in favor of A, and then the post-deadline beacon resolved the tie in favor of B. The final statement had committed to the wrong winner.

## Root cause

The guard for adding a statement only checks `closed?`, which is purely deadline-based:

- `app/services/api_helper.rb:1011` — `raise "Decision must be closed to add a statement" unless decision.closed?`
- `app/controllers/decisions_controller.rb:503` — same `closed?` check for the HTML path
- `app/models/application_record.rb:144` — `closed?` is just `deadline <= Time.now`

Meanwhile the tiebreaker beacon is drawn separately and later:

- `app/jobs/lottery_draw_job.rb:24` — if the deadline is still in the future, reschedule for `deadline + 5.seconds`; the target drand round isn't published until after the deadline
- Beacon presence is tracked by `Decision#beacon_drawn?` (`lottery_beacon_round.present?`) — `app/models/decision.rb:56`
- `LotteryDrawJob` runs for both `is_lottery?` and `is_vote?` decisions (`app/services/api_helper.rb:1002`)

So `closed? == true` but `beacon_drawn? == false` is a real, reachable state for any vote/lottery decision, for as long as it takes the job to run and the beacon round to publish.

## Proposed fix

Add a "fully resolved" notion and require it (not just `closed?`) before a final statement can be created:

- For `is_lottery?` / `is_vote?` decisions: require `closed?` **and** `beacon_drawn?` (results computed from the resolved beacon).
- For `is_executive?` decisions: no beacon is involved, so `closed?` remains sufficient.

Suggested shape: a `Decision#fully_resolved?` predicate, used by both the API (`api_helper.rb#add_statement`) and the controller (`decisions_controller.rb#add_statement_action`), with a clear error message (e.g. "Decision results are not yet final — the tiebreaker beacon has not been drawn yet."). The UI should also hide/disable the write-statement affordance until `fully_resolved?` (see `app/views/decisions/show.html.erb:161`).

## Notes

- This is a stricter superset of the current rule: anything that fails `closed?` still fails; we additionally block the closed-but-unresolved window.
- Worth a test covering the closed-but-beacon-not-yet-drawn state for a tied vote decision.

melody-harmonic:
Closes #267.

## Problem

A decision's final statement could be written as soon as the **deadline** passed, even though for vote/lottery decisions the winner of a tie isn't known until the drand tiebreaker **beacon** is drawn — and that draw happens asynchronously, a few seconds *after* the deadline (`LotteryDrawJob`). That opened a window where a statement could commit to a provisional winner that the beacon later overturned. This actually bit us: a tie's provisional sort showed A on top, a statement closed in favor of A, then the beacon resolved the tie for B.

The only guard was `closed?` (deadline-based), at `api_helper.rb#add_statement` and `decisions_controller.rb#add_statement_action`.

## Fix

Add `Decision#fully_resolved?` — `closed?` **and**, for vote/lottery decisions, `beacon_drawn?` — and require it (not just `closed?`) before a final statement can be created. Executive decisions involve no beacon, so `closed?` stays sufficient for them.

- **`Decision#fully_resolved?`** + `requires_beacon?` + a shared `UNRESOLVED_STATEMENT_ERROR` message.
- **Both guards** (API + controller) gain a second check after the existing `closed?` one, so the *not-closed* and *closed-but-unresolved* cases get distinct, accurate messages. The markdown path returns **409 Conflict**.
- **Show page** swaps the write-statement form for a brief "results aren't final yet — the beacon hasn't been drawn" notice during the window.
- **`add_statement` action description** and the **decisions / lottery help docs** now state the beacon requirement.

`LotteryDrawJob` draws a beacon for *every* closed vote/lottery decision (tie or not), so this never permanently blocks — it only closes the brief async window. It's a strict superset of the old rule: anything that failed `closed?` still fails.

## Tests

- **Model**: `fully_resolved?` across open/closed × vote/lottery/executive, with and without a drawn beacon.
- **Controller**: a closed vote with no beacon is blocked (HTML + a 409 on the markdown path); once `lottery_beacon_round` is set the same decision accepts the statement. Two existing tests that closed a vote and expected the statement to save now also set `lottery_beacon_round`.

⚠️ Tests are red-green by construction but **unverified locally** (no Docker in this environment) — leaning on CI.

### DIFF
--- a/app/controllers/decisions_controller.rb
+++ b/app/controllers/decisions_controller.rb
@@ -508,6 +508,14 @@ class DecisionsController < ApplicationController
       return
     end
 
+    unless @decision.fully_resolved?
+      respond_to do |format|
+        format.html { redirect_to @decision.path, alert: Decision::UNRESOLVED_STATEMENT_ERROR }
+        format.md { render_action_error({ action_name: 'add_statement', resource: @decision, error: Decision::UNRESOLVED_STATEMENT_ERROR, status: :conflict }) }
+      end
+      return
+    end
+
     begin
       api_helper(params: { text: params[:text] || params[:final_statement] }).add_statement
       respond_to do |format|
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -62,6 +62,30 @@ class Decision < ApplicationRecord
     is_lottery? && beacon_drawn?
   end
 
+  sig { returns(T::Boolean) }
+  def requires_beacon?
+    is_lottery? || is_vote?
+  end
+
+  # Error surfaced when a final statement is attempted on a decision that is
+  # closed but not yet fully resolved (the tiebreaker beacon has not been
+  # drawn). See #fully_resolved? and issue #267.
+  UNRESOLVED_STATEMENT_ERROR = "Decision results are not yet final — the tiebreaker beacon has not been drawn yet."
+
+  # A decision is fully resolved once voting has closed AND — for vote/lottery
+  # decisions — the drand tiebreaker beacon has been drawn. The beacon is drawn
+  # asynchronously *after* the deadline (see LotteryDrawJob), so `closed?` alone
+  # is true during a window in which a tied result can still flip. Final
+  # statements must wait for full resolution so they can't commit to a
+  # provisional winner that the beacon later overturns. Executive decisions
+  # involve no beacon, so `closed?` is sufficient. (#267)
+  sig { returns(T::Boolean) }
+  def fully_resolved?
+    return false unless closed?
+
+    requires_beacon? ? beacon_drawn? : true
+  end
+
   AUDIT_CHAIN_LAUNCH_DATE = Time.utc(2026, 5, 5)
 
   sig { returns(T::Boolean) }
--- a/app/services/actions_helper.rb
+++ b/app/services/actions_helper.rb
@@ -427,7 +427,7 @@ class ActionsHelper
       visibility: :by_collective,
     },
     "add_statement" => {
-      description: "Add or update the final statement on this decision. Only available after the decision is closed.",
+      description: "Add or update the final statement on this decision. Only available once the decision is fully resolved — closed, and for vote/lottery decisions the tiebreaker beacon has been drawn.",
       params_string: "(text)",
       params: [
         { name: "text", type: "string", description: "The statement text" },
--- a/app/services/api_helper.rb
+++ b/app/services/api_helper.rb
@@ -1013,6 +1013,7 @@ class ApiHelper
     raise "Unauthorized" unless decision.can_write_statement?(current_user)
 
     raise "Decision must be closed to add a statement" unless decision.closed?
+    raise Decision::UNRESOLVED_STATEMENT_ERROR unless decision.fully_resolved?
 
     statement = create_or_update_statement!(decision, params[:text])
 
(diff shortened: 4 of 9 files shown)


melody-harmonic:

**Verdict: Changes requested** — one blocking correctness gap that defeats the PR's own goal.

### Blocking

**`ApiHelper#close_decision` writes an inline `final_statement` before the beacon is drawn — an unguarded bypass of exactly what #267 prohibits.**

`add_statement` is now correctly guarded in all three surfaces (API helper, controller HTML, controller markdown → 409). But `create_or_update_statement!` has a *second* call site that I missed: the inline `final_statement` accepted by `close_decision`.

```ruby
# app/services/api_helper.rb:995
create_or_update_statement!(decision, params[:final_statement]) if params[:final_statement].present?
# ...
LotteryDrawJob.perform_later(decision.id) if decision.is_lottery? || decision.is_vote?
```

The statement is persisted inside the close transaction, while the beacon is only *enqueued* (`perform_later`) afterward — so at write time `fully_resolved?` is necessarily `false` for a vote/lottery decision. A creator can commit a final statement onto a still-provisional winner just by passing `final_statement` on close, which is the precise scenario this PR set out to block. Reachable from the controller's `close_decision_action` (passes `final_statement:` straight through) on both HTML and markdown.

The retained test `test/controllers/decisions_controller_test.rb:302` ("creator can close decision with final statement") proves the hole — it closes a vote decision and asserts the statement is saved with no beacon drawn.

**Fix:** put the guard in `create_or_update_statement!` itself so every path is covered (cleaner than guarding each call site), or add the same `fully_resolved?` check in `close_decision`. The line-302 test then needs to flip to asserting the statement is rejected/deferred, and a new test should cover close-with-inline-statement on an unresolved vote/lottery.

**Doc follow-on:** `app/views/help/decisions.md.erb` now says "When closing an executive decision, the creator can optionally include the final statement as part of the close action" — implying close-time statements are executive-only, but the code still accepts and persists them for vote/lottery. Reconcile once the guard lands.

### What's right

- `fully_resolved?` is correct and well-matched to the codebase: vote and lottery both always get a drand beacon after the deadline, executive never does, and `_results.html.erb` already gates winner highlighting on the identical `closed? && beacon_drawn?` condition. Good reuse of the existing concept.
- The `add_statement` guard is complete across API + both controller formats, with the markdown path returning a correct 409.

### Test assessment

Model tests cover the predicate well (open/closed-no-beacon/closed-with-beacon for vote, lottery transition, executive-no-beacon). Controller tests exercise the `add_statement` guard (HTML stays nil, markdown 409, resolved→allows) and correctly update the two pre-existing happy-path tests to set a beacon round. The gap mirrors the blocking issue: no test for the `close_decision` inline-statement path.

Issue #279, opened 2026-06-28
table ux improvements


danallison:
the ux for table notes has room for improvement. ideally it would work like a typical spreadsheet app. maybe there are some open source options that would work out of the box. however, we must make sure that all spreadsheet-based calculations happen on the backend only so that the markdown view and functionality matches the html. no frontend calculations.

the baseline improvement is to render the table correctly with the markdown renderer. currently table contents in the html view are just raw text, which is annoying when cells contain links that you want to click on.

melody-harmonic:
**Investigation (melody)**

**How table notes work today:** table content lives in a `table_data` JSON column on `Note` (migration `20260427205555_add_table_data_to_notes.rb`), shaped `{ description, columns, rows }`. `NoteTableService` (`app/services/note_table_service.rb`) owns all mutations/queries, and `NoteTableFormatter.to_markdown` writes a markdown rendition into `note.text` so the `.md` view stays consistent.

**Confirmed bug (baseline issue):** in `app/views/notes/_table.html.erb`, the description is rendered through markdown (`<%= markdown(table.description) %>`) but cells are plain interpolation — `<td><%= row[col["name"]] %></td>` (~line 28). So links in cells render as raw text in HTML.

**Baseline fix is small:** run cell values through the existing `markdown()` helper (which uses `MarkdownRenderer`, Redcarpet, with table tags already allowed). The `.md` view already shows the markdown form, so this just makes HTML match. Worth confirming sanitization/escaping behavior on cell content.

**Backend-only calc note:** aggregation already lives server-side in `NoteTableService` (`query_rows`, `summarize`: count/sum/avg/min/max), consistent with the "no frontend calculations" requirement. The "spreadsheet app" goal is the larger follow-on; the baseline markdown-rendering fix is independent and cheap.

PR #305, opened 2026-06-30
Fix #286: edit table rows in place in the HTML UI

### LINKED ISSUE #286: no ability to update table note rows in html ui
rows can currently only be deleted and created in the html ui, but not updated

melody-harmonic:
Closes #286 (sub-issue of #279).

## Problem

Table notes could **create** and **delete** rows in the HTML UI, but not **update** them — even though the `update_row` backend already existed (`NoteTableService#update_row!`, `ApiHelper#update_row`, and the `execute_update_row` route/action). The HTML view just never offered an edit affordance, and `execute_update_row` only spoke markdown.

## Fix

In-place row editing, mirroring the existing add/delete-row form pattern (plain forms + Turbo, no fetch):

- **`_table.html.erb`** — each editable row gets an **Edit** button that swaps its cells' rendered values for text inputs and reveals **Save**/**Cancel**. The cell inputs are associated (via the HTML `form=` attribute) with **one hidden `update_row` form per row**, rendered just after the table (a `<form>` can't sit between `<tr>`s). Save submits that form through Turbo → 302 redirect → the page reloads with the new values.
- **`table_row_edit_controller.ts`** — a small Stimulus controller (`table-row-edit`) toggling each row between display and edit. **Cancel** restores the server-rendered values (`input.defaultValue`) and exits edit mode.
- **`NotesController#execute_update_row`** — adds an HTML branch (redirect with a flash notice/alert), matching `add_row`/`delete_row`. The markdown branch is unchanged, so agent dispatch keeps its structured action result.

No new routes, action definitions, or capability-check entries — `update_row` already existed end-to-end; this is the missing HTML surface.

## Tests

- **Controller**: `update_row` via HTML form redirects and persists the new cell values; a missing `row_id` redirects with a "not found" alert; the table show page renders the per-row update form, the `data-controller` row, the wired inputs, and the edit/save buttons for an editor.
- **JS (vitest)**: the controller toggles display/inputs/buttons on Edit, and Cancel discards edits and restores display mode.

⚠️ Red-green by construction but **unverified locally** (no Docker/node in this environment) — leaning on CI.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -385,9 +385,15 @@ class NotesController < ApplicationController
 
   def execute_update_row
     api_helper.update_row
-    render_action_success({ action_name: "update_row", resource: current_note, result: "Row updated." })
+    respond_to do |format|
+      format.html { redirect_to current_note.path, notice: "Row updated." }
+      format.md { render_action_success({ action_name: "update_row", resource: current_note, result: "Row updated." }) }
+    end
   rescue RuntimeError, ActiveRecord::RecordInvalid => e
-    render_action_error({ action_name: "update_row", resource: current_note, error: e.message })
+    respond_to do |format|
+      format.html { redirect_to current_note.path, alert: e.message }
+      format.md { render_action_error({ action_name: "update_row", resource: current_note, error: e.message }) }
+    end
   end
 
   def describe_delete_row
--- a/app/javascript/controllers/index.ts
+++ b/app/javascript/controllers/index.ts
@@ -80,6 +80,7 @@ import CommitmentSubtypeController from "./commitment_subtype_controller"
 import DecisionSubtypeController from "./decision_subtype_controller"
 import DialogController from "./dialog_controller"
 import TabsController from "./tabs_controller"
+import TableRowEditController from "./table_row_edit_controller"
 import WebhookTestController from "./webhook_test_controller"
 
 // Register all controllers
@@ -152,6 +153,7 @@ application.register("kebab-menu", KebabMenuController)
 application.register("summary-toggle", SummaryToggleController)
 application.register("dialog", DialogController)
 application.register("tabs", TabsController)
+application.register("table-row-edit", TableRowEditController)
 application.register("webhook-test", WebhookTestController)
 
 export { application }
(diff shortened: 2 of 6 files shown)


melody-harmonic:

**Verdict: Approve with nits** — solid, no blocking issues.

### What's right

- Correctly **reuses** the existing `api_helper.update_row` → `NoteTableService#update_row!` path; the only controller change is adding the `respond_to`/`format.html` branch, mirroring `add_row`/`delete_row` (PR #284). No duplicated logic.
- **Authorization** enforced via the same `authorize_content_edit!` gate as add/delete (`api_helper.rb:439`) — not "any user can edit any table."
- **CSRF**: uses `form_with` (auto token) submitted through Turbo; no hand-rolled fetch. Session requests stay protected.
- **Arbitrary-key writes blocked**: `update_row!` only writes keys present in `column_names` (`note_table_service.rb:97`); `_harmonic_*` and unknown keys can't be set.
- **XSS**: edited cell content renders via `markdown_inline` → `MarkdownRenderer.sanitize`; the raw value in the input is Rails-auto-escaped; the error flash (which interpolates the user-supplied `row_id`) is also rendered through `markdown_inline`, so it's escaped at render time.

### Concerns

- **No authorization-failure controller test.** Auth is enforced and the edit affordance only renders when `can_edit`, so this is a coverage gap, not a live hole — but worth adding a non-owner-edits-owner-note case asserting rejection. Also no test that an invalid/unknown column is silently dropped (that filter is the security-relevant guard, untested).
- **Concurrent-edit lost-update.** `update_row!` does a read-modify-write of the whole `table_data` blob with no version check (inherited from #284). In-place editing makes simultaneous edits much more likely → silent last-writer-wins. A Save against a server-deleted row correctly raises and flashes. Worth a follow-up, not a blocker.

### Nits

- **Per-row hidden form scales O(rows)** (`_table.html.erb:93`): one `<form>` per row, each duplicating the CSRF token. Fine for typical table-note sizes; a single shared form with a JS-populated `row_id` would avoid the multiplication. (The comment correctly explains why forms must live outside `<table>`.)
- **No Escape-to-cancel keybinding** in the Stimulus controller (click-only). Enter submits via the linked form (reasonable). Minor a11y gap.
- On failure the in-progress edits are discarded and the error shows only in the top-of-page flash, not tied to the row. Acceptable for a first cut.

### Test assessment

Controller tests cover happy path (asserts actual data mutation after reload), missing-row (redirect + flash), and the rendered affordance (per-row form, `data-controller`, wired inputs, buttons). The TS test genuinely exercises display/edit/cancel DOM state but never submits — acceptable, since submission is plain HTML form behavior. Add: auth-failure and invalid-column cases.

PR #306, opened 2026-06-30
Fix #299: don't notify @handles inside code spans/blocks

### LINKED ISSUE #299: Notification mention-parser pings on @handles inside code spans/blocks
## Summary

The **notification** mention-parser does not share the **renderer's** code-span exclusion. An `@handle` written inside an inline code span or fenced code block still generates a real mention notification, even though it renders as plain (non-linked) text.

Concretely: posting a note whose body contained `` `@trio` `` (inside a code span, used as a documentation example) fired a real "Trio isn't enabled" notification.

## Why the two diverge

PR #295 (merged + deployed 2026-06-29) made the Markdown renderer link `@handle` mentions and skip them inside code spans/blocks/links — but it did this in the rendering layer (a Redcarpet `normal_text` renderer), which only affects HTML output. The notification dispatch path is a separate parser and was intentionally untouched by #295.

Notification dispatch resolves mentions via `MentionParser.parse_for_notification` → `MentionParser.parse` → `MentionParser.extract_handles`, which runs a flat regex (`MENTION_PATTERN = /@([a-zA-Z0-9_-]+)/`) over the raw Markdown source with no awareness of code spans, code blocks, or link syntax. So any `@handle` anywhere in the text — including inside `` `code` `` — is treated as a mention for notification purposes.

- `app/services/mention_parser.rb` — `extract_handles` / `parse` / `parse_for_notification`
- `app/services/notification_dispatcher.rb` — caller

## Consequence

Code samples, docs, and any prose that quotes a handle inside code formatting keep generating spurious mention notifications.

## Proposed fix

Align the notification parser's handle extraction with the renderer's #295 exclusion logic, so mentions inside code spans/blocks (and link targets/text, matching the renderer) are skipped before dispatch. Options:

1. Strip/skip code spans + fenced blocks in `extract_handles` (or a shared pre-pass) before applying `MENTION_PATTERN`, so both `parse` (notifications) and `resolve_paths` use the same exclusion.
2. Drive notification extraction through the same Redcarpet `normal_text` pass the renderer uses, collecting handles from non-code/non-link text only, to keep a single source of truth.

Either way the goal is one mention-detection rule shared by rendering and notification dispatch.

## Notes

- Renderer fix reference: #295.
- No implementation requested yet — filing for tracking per @danallison.

melody-harmonic:
Closes #299.

## Problem

The **notification** mention-parser didn't share the **renderer's** code-span exclusion. An `@handle` inside an inline code span or fenced code block still generated a real mention notification, even though it renders as plain (non-linked) text. Concretely, a note containing `` `@trio` `` (a documentation example) fired a real "Trio isn't enabled" notification.

PR #295 fixed the renderer by linkifying mentions in a Redcarpet `normal_text` callback — which gets code-span exclusion for free, because Redcarpet routes code through callbacks that never reach `normal_text`. But notification dispatch is a separate parser (`MentionParser.parse` → `extract_handles`) that ran a flat regex over the raw Markdown source.

## Fix

Strip fenced code blocks and inline code spans in the **shared** `MentionParser.extract_handles`, before applying `MENTION_PATTERN`. Both paths funnel through `extract_handles`:
- **notifications**: `parse` / `parse_for_notification` (over raw text) — now correctly excludes code.
- **rendering**: `resolve_paths` (called from the renderer's `normal_text`) — only ever receives a single, already code-free text node, so stripping is a **no-op** and renderer behavior is unchanged.

One mention-detection rule, shared.

**Scope**: covers code spans and fenced (```` ``` ```` / `~~~`) blocks — the reported case. Handles in link *targets* are out of scope: the renderer resolves mentions in link *text* too, so full link alignment is a separate, subtler concern worth its own change.

## Tests

`extract_handles` skips handles in inline spans and fenced/tilde blocks, still returns real handles alongside a code-span example and adjacent to a code span, and isn't tripped by a lone unbalanced backtick; `parse` doesn't notify a resolvable handle that only appears inside a code span. The renderer's existing code-span tests are unaffected (different path).

⚠️ Red-green by construction but **unverified locally** (no Docker here) — leaning on CI.

### DIFF
--- a/app/services/mention_parser.rb
+++ b/app/services/mention_parser.rb
@@ -1,18 +1,48 @@
 # typed: true
 
+require "redcarpet/render_strip"
+
 class MentionParser
   extend T::Sig
 
   MENTION_PATTERN = /@([a-zA-Z0-9_-]+)/
   TRIO_HANDLE = "trio"
 
-  # A fenced code block: an opening run of >=3 backticks or tildes at line
-  # start, through the matching closing fence (or end of text if unclosed).
-  FENCED_CODE_BLOCK = /^[ \t]*(`{3,}|~{3,})[^\n]*\n.*?(?:^[ \t]*\1[ \t]*$|\z)/m
+  # A Redcarpet plain-text renderer that drops code while keeping every other
+  # kind of text. It subclasses StripDown — which already extracts the text
+  # content of each node — and blanks only the code callbacks. The upshot is
+  # that "what counts as code" is decided by the same Markdown tokenizer the
+  # HTML renderer uses (MarkdownRenderer, #295), so the notification path can't
+  # drift from the render path the way a separate hand-rolled regex would: a
+  # fenced block, an *indented* (4-space/tab) block, an inline span, and a
+  # mention inside a nested list are all classified here exactly as on render.
+  # Code is replaced with whitespace (never nothing) so text on either side of
+  # a span can't fuse into a spurious handle — "@" `x` "foo" must not read as
+  # "@foo". (#299)
+  class CodeStrippingRenderer < Redcarpet::Render::StripDown
+    def block_code(_code, _language)
+      "\n"
+    end
 
-  # An inline code span: a run of backticks, the shortest span up to the same
-  # run. (`@trio` -> matched and excluded; a lone stray backtick won't match.)
-  INLINE_CODE_SPAN = /(`+)[^`]*?\1/
+    def codespan(_code)
+      " "
+    end
+  end
+
+  # Tokenizer extensions mirror MarkdownRenderer's so code is recognized
+  # identically on both paths: fenced_code_blocks so ``` fences route through
+  # block_code, and no_intra_emphasis so underscores in handles aren't eaten as
+  # emphasis (e.g. @a_b_c stays one handle rather than splitting on the _b_).
+  CODE_STRIPPER = T.let(
+    Redcarpet::Markdown.new(
+      CodeStrippingRenderer,
+      autolink: true,
+      tables: true,
+      fenced_code_blocks: true,
+      no_intra_emphasis: true,
+    ),
+    Redcarpet::Markdown,
+  )
 
   sig do
     params(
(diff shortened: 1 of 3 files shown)


melody-harmonic:

**Verdict: Approve with nits** — core fix is correct and fail-safe; the gaps are scope-vs-issue divergences worth deciding on explicitly.

### What's right

- Inline code spans and ```` ``` ````/`~~~` fenced blocks are stripped before handle extraction. Real mentions adjacent to code still notify; the happy path is preserved.
- **Fail-safe on malformed input**: unclosed fences/backticks swallow to end-of-text (via the `\z` fallback and the `[^`]` guard) rather than dropping the document or leaking. The failure direction is under-notify, not over-notify — the right choice.

### Concerns

1. **Indented (4-space) code blocks are not stripped.** Only fenced + inline patterns are defined (`mention_parser.rb:11`). A handle in a 4-space-indented block (`    @alice ...`) still notifies, while Redcarpet renders it as plain code text — the same #299 bug via a different construct. The issue says "code spans/**blocks**" and asks for one shared rule. Either handle indented blocks or document them as out of scope in a code comment.
2. **Second, divergent definition of "what is code."** The renderer uses Redcarpet's real tokenizer (`MentionRenderer#normal_text`); this PR re-derives code detection with regex. The two already disagree on multi-backtick spans with an interior backtick (CommonMark = one span; the regex pairs differently). Outcomes match on the cases probed, but this is the drift risk the issue flagged (it leaned toward driving notification extraction through the same Redcarpet pass). At minimum, add a comment that any renderer change must be mirrored here.
3. **Mild O(n²) backtracking** on long backtick runs: `INLINE_CODE_SPAN = /(`+)[^`]*?\1/` backtracks quadratically (~50k backticks ≈ multiple seconds). Runs on the user-supplied notification path. Low severity (large input needed), trivially mitigated with a possessive quantifier or a length guard.

### Nits

- Pre-existing parser/renderer divergences left in place (out of scope, worth a mention): `someone@example.com` still extracts `example`; `@handle` inside link targets isn't aligned with the renderer.
- Fence close uses exact-length `\1`; CommonMark allows a longer closing fence (`>= ` opening). Mismatches fall through to `\z` (safe under-notify). Edge-casey, fine to leave.

### Test assessment

Good coverage of the chosen scope: inline span, backtick + tilde fenced blocks, still-notifies-alongside-a-span, handle adjacent to a span, lone unbalanced backtick, and integration via `parse`. Missing cases that map to the concerns: **indented code block** (would currently leak — this gap hides concern #1), multi-backtick span with interior backtick (advertised in a comment but untested), unclosed fenced block (lock in the drop-to-EOF contract), and a ReDoS-input guard.

PR #308, opened 2026-06-30
Fix #270: MCP tool-call log page for agent principals

### LINKED ISSUE #270: MCP tool call log
All mcp tool calls are logged in the database, but there's currently nowhere that an agent's principal can view the logs. We've talked about unifying the agent task run model for both internal and external agents, but that's a longer term plan. Initially we just need a page where users can inspect the logs.

melody-harmonic:
Closes #270.

## Problem
Every MCP tool call is already persisted to `mcp_tool_call_logs`, but there was nowhere for an agent's principal to actually read them.

## What this adds
A principal-facing **list + detail** view of an agent's MCP tool calls, living under the agent and mirroring the existing `runs` / `show_run` pattern:

- **Routes**
  - `GET /ai-agents/:handle/mcp-tool-calls` — list
  - `GET /ai-agents/:handle/mcp-tool-calls/:log_id` — detail
- **Controller** (`AiAgentsController`)
  - `#mcp_tool_calls` — most-recent-first, capped at `MCP_TOOL_CALLS_PER_PAGE` (100) since the table is append-only with no retention policy yet
  - `#show_mcp_tool_call` — single call with full args/context/resources
  - Both are human-only and scoped to the caller's **own** agents via `find_ai_agent_by_handle`, plus the model's tenant default scope
- **Views** — HTML (pulse styling) and Markdown for both list and detail. Detail shows status, duration, source (internal task-run vs external client), timestamp, request id, arguments, context, and any resources the call touched.
- **Model** — `McpToolCallLog.recent` scope and `internal?` / `source_label` helpers.
- **Navigation** — "MCP Tool Calls" link added to the agent show page (HTML + Markdown).

## Scoping notes
- Works for **external** agents (Claude Desktop/Code/Cursor) as well as internal task-run agents — the source column distinguishes them.
- This is the read-only "initial" page the issue asks for; the longer-term unified agent-task-run model is out of scope here.

## Tests
Controller integration tests (access, tenant/owner scoping, 404s for unknown agent/log and cross-agent logs, feature-flag gating, markdown rendering) and model tests (`recent` ordering, `internal?`/`source_label`).

## Caveat
Written red-green by construction but **unverified locally** — this environment has no Docker, and the repo's tests/rubocop/sorbet run in containers. Relying on CI.

### DIFF
--- a/app/controllers/ai_agents_controller.rb
+++ b/app/controllers/ai_agents_controller.rb
@@ -6,13 +6,20 @@ class AiAgentsController < ApplicationController
 
   TASK_RUNS_PER_MINUTE = 5
 
+  # Cap the MCP tool-call log page at the most recent N rows. The table is
+  # append-only with no retention policy yet (see McpToolCallLog), so an
+  # unbounded query would grow without limit for a chatty agent.
+  MCP_TOOL_CALLS_PER_PAGE = 100
+
   before_action :set_sidebar_mode,
-                only: [:new, :index, :show, :settings, :run_task, :execute_task, :runs, :show_run, :cancel_run, :create, :execute_create_ai_agent, :deactivate,
+                only: [:new, :index, :show, :settings, :run_task, :execute_task, :runs, :show_run, :cancel_run, :mcp_tool_calls, :show_mcp_tool_call, :create,
+                       :execute_create_ai_agent, :deactivate,
                        :reactivate,]
   before_action :require_any_ai_agents_enabled, only: [
     :index, :show, :settings, :update_settings,
     :describe_update_ai_agent, :execute_update_ai_agent, :settings_actions_index,
     :describe_update_notification_preferences, :execute_update_notification_preferences,
+    :mcp_tool_calls, :show_mcp_tool_call,
   ]
   before_action :require_internal_ai_agents_enabled, only: [:run_task, :execute_task, :runs, :show_run, :cancel_run]
   before_action :require_flag_for_create_mode, only: [:new, :create, :execute_create_ai_agent]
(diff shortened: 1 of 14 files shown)


melody-harmonic:

**Verdict: Approve with nits** — authorization is sound, IDOR structurally prevented; only cleanups remain.

### What's right

- **Authorization (the main risk) is correct.** Both actions resolve the agent via `find_ai_agent_by_handle`, which constrains the lookup to `current_user.ai_agents` at the DB level; logs are then queried with `where(user: @ai_agent)`. A user can only ever reach logs for agents they're the principal of. The show action additionally re-scopes the log lookup by agent, so passing another principal's handle simply 404s.
- Faithfully mirrors the existing `runs`/`show_run` convention (controller shape, 403/404 handling, route nesting, view structure). JSON blobs are auto-escaped in the HTML views.

### Concerns

- **Wasted eager-load** (`ai_agents_controller.rb:368`): `.includes(:api_token, :ai_agent_task_run)` loads associations the index view never reads (`source_label`/`internal?` only read the `ai_agent_task_run_id` *column*; `api_token` is never referenced). Not an N+1 — extra preload queries. Drop the `includes`.
- **`display_path` rendered as an unsanitized `href`** (`show_mcp_tool_call.html.erb:48`). This is **identical to the pre-existing `show_run.html.erb:156` pattern**, so not introduced here — but if these stored paths aren't validated at write time, a `javascript:` URI would be a stored-XSS vector across both pages. Worth a follow-up to validate at write time; out of scope for this PR.

### Nits

- **Non-deterministic ordering** (`mcp_tool_call_log.rb:40`): `order(created_at: :desc)` has no secondary key; same-millisecond logs sort unstably. Matches `AiAgentTaskRun.recent`, but `order(created_at: :desc, id: :desc)` would be stable.
- **"Pagination" is a hard cap** (`MCP_TOOL_CALLS_PER_PAGE`, `.limit(100)` with no offset/cursor): logs beyond the newest 100 are unreachable. The cap is reasonable and the view copy is honest ("max 100"), but `_LIMIT` would name it better.
- **Markdown pipe-injection** in `.md.erb` table cells: a `tool_name`/path containing `|` breaks table rendering. Cosmetic, no security impact; values are effectively server-controlled enums.

### Test assessment

Strong. The critical authorization-denied cases are present: "user cannot view another user's agent MCP tool calls" → `:not_found`, and "detail returns 404 for a log belonging to a different agent" → confirms the `where(user: @ai_agent)` scoping (the right IDOR probe). Also covers unauthenticated redirect, non-existent handle 404, unknown log id 404 (UUID, so no 500), feature-flag-disabled 403, empty state, and Markdown content negotiation. Model tests cover `recent` ordering and the `internal?`/`source_label` branch. Minor gap: no explicit test that a non-human (ai_agent) user hits the `human?` 403 guard (duplicated from the well-tested `runs` actions).

danallison:
seems like the most useful information gets left out, and the table contains the least interesting (and most repetitive) info. the table should show the timestamp, tool, path, action name, and intention. we'll have to do another design pass on this later, but for now let's just at least include the most useful info.

melody-harmonic:
Done (46f90694, pushed). The list table now leads with the useful columns — **timestamp · tool · path · action · intention** — and drops status/duration/source, which were the repetitive ones. Non-applicable cells (e.g. no action for a `fetch_page`, no path for `search`) render an em dash. Path/action come from `arguments`, intention from the `context` block, via new null-safe `McpToolCallLog` accessors; the Markdown list mirrors the same columns. Full status/duration/source stays on the per-call detail page. Agreed a proper design pass comes later — this is just the quick win. CI-verified but not locally test-run on my end.

PR #311, opened 2026-06-30
Add docker-compose.test.yml for running tests locally

melody-harmonic:
## What

Adds `docker-compose.test.yml` so the Rails test suite can be run locally in Docker, plus a short "Running tests locally" note in the README.

Follow-up to Dan's note about getting tests running on my machine ([harmonic-devs note `9f9fc34e`](https://www.harmonic.social/collectives/harmonic-devs/n/9f9fc34e)).

## Design

A **trimmed, CI-faithful** stack — just `test` + `db` + `redis`:

- **Mirrors CI** (`.github/workflows/ci.yml` → `rails-tests`): `postgres:13`, `redis:6.2-alpine`, the same env vars (`RAILS_ENV=test`, `DB_HOST`, `AUTH_MODE=oauth`, `HOSTNAME`, `PRIMARY_SUBDOMAIN`, `AUTH_SUBDOMAIN`, `REDIS_URL`), and `db:create` + `db:schema:load` before `rails test`.
- **Drops** everything tests don't need — web/js/sidekiq, caddy/ngrok, mailcatcher, and the 512 MB-each `litellm`/`agent-runner` — to keep memory/disk down (the box this was requested for is small).
- **No source bind-mount.** The `Dockerfile` already bakes app code, gems, `node_modules` and built JS assets via `COPY . .` + `npm run build`, and runs as `harmonic` (uid 1000). Bind-mounting `.:/app` would hide those, so the file relies on the baked image instead — rebuild after changing code. (This also matches CI's fresh-checkout semantics.)
- **db healthcheck** (`pg_isready`) gates the test runner via `depends_on: condition: service_healthy`, avoiding a `db:create` race on startup.

## Usage

```bash
docker compose -f docker-compose.test.yml build
docker compose -f docker-compose.test.yml run --rm test                                   # full suite
docker compose -f docker-compose.test.yml run --rm test bundle exec rails test test/models  # a subset
docker compose -f docker-compose.test.yml down -v
```

## ⚠️ Unverified

I can't run this yet — my current box has no Docker and is at 98% disk / ~458 MB RAM (the very constraint that motivated the note). The compose file was written against `main`'s `Dockerfile`, `docker-compose.yml`, `config/database.yml`, and `ci.yml`, but it has **not** been executed. Worth a real run on the upgraded instance before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,21 @@ cp .env.example .env
 ./scripts/stop.sh    # stop the app
 ```
 
+## Running tests locally
+
+The Rails test suite runs in Docker via a trimmed, CI-faithful compose file
+([`docker-compose.test.yml`](docker-compose.test.yml)) — just a test runner plus
+Postgres and Redis, matching the versions and environment used in CI.
+
+```bash
+docker compose -f docker-compose.test.yml build
+docker compose -f docker-compose.test.yml run --rm test                                   # full suite
+docker compose -f docker-compose.test.yml run --rm test bundle exec rails test test/models  # a subset
+docker compose -f docker-compose.test.yml down -v                                         # tear down
+```
+
+The image bakes the app code, gems and assets, so rebuild after changing code.
+
 ## License
 
 MIT
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ cp .env.example .env
 
 ## Running tests locally
 
+### Rails test suite (Docker)
+
 The Rails test suite runs in Docker via a trimmed, CI-faithful compose file
 ([`docker-compose.test.yml`](docker-compose.test.yml)) — just a test runner plus
 Postgres and Redis, matching the versions and environment used in CI.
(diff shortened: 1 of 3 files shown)


danallison:
what about typescript tests? if you're working on `agent-runner` or `harmonic-bridge` or just the frontend ui typescript, how does that work with this setup?

melody-harmonic:
Good question — and the answer is that **none of the TypeScript touches this compose file at all.** `docker-compose.test.yml` exists purely for the Rails suite, because that's the only thing that needs Postgres + Redis up. The TS tests are plain Node and run natively.

That mirrors CI exactly: there's a separate `typescript` job (`.github/workflows/ci.yml`) that is just `setup-node@v4` → `npm install` → `npm test`. No services, no containers. It covers three independent packages:

- **Frontend UI (root `package.json`)** — `npm run typecheck` + `npm test` (vitest + jsdom over `app/javascript/**/*.test.ts`) + `npm run build` (esbuild). jsdom fakes the DOM, so no browser and no server.
- **`agent-runner/`** — `cd agent-runner && npm install && npm run build && npm run typecheck && npm test` (vitest).
- **`harmonic-bridge/`** — `cd harmonic-bridge && npm install && npm run typecheck && npm test` (node's built-in `--test` runner) + `npm run build`.

So for frontend Stimulus work, `agent-runner`, or `harmonic-bridge`, you **don't need the upgraded box or Docker at all** — Node 22 is already installed here, and these run in seconds. The only thing stopping them on the *current* box is the same disk wall as everything else (179 MB free, 98% full): the root `node_modules` won't fit. `agent-runner`/`harmonic-bridge` have tiny dep trees, but the box is full regardless. Once we have disk headroom, all three run with no containers.

**The one exception is Playwright e2e** (`npm run test:e2e`, the `e2e/` dir + `playwright.config.ts`). Those drive a real Chromium against a *running* app, so they need two things the unit tests don't:
1. The browser: `npm run playwright:install` (chromium, ~adds a few hundred MB).
2. The actual app reachable at `E2E_BASE_URL` (default `https://app.harmonic.local`) — i.e. web + db + redis up, plus the authed storage state from `e2e/global-setup.ts`.

So e2e is the only TS surface that needs the app stack running (Docker, or pointed at a deployed instance). Worth noting CI doesn't currently run Playwright in the `typescript` job — it's there for local/manual use.

**Net for sizing:** the 8 GB / 4 vCPU target still covers everything, including e2e. Nothing about TS changes the floor — if anything the unit-test TS is the *lightest* thing we'd run and could've lived on a much smaller box on its own.

If useful, I can extend the "Running tests locally" README note in this PR to document the npm/vitest commands (and the e2e prerequisites) alongside the Rails docker path — right now it only covers the Ruby suite. Want me to add that?

danallison:
yes

melody-harmonic:
Done — pushed `16ac66c7`. The "Running tests locally" section now has three subsections:

- **Rails test suite (Docker)** — the existing `docker-compose.test.yml` path.
- **TypeScript / JavaScript tests (no Docker)** — the npm/vitest commands for the frontend UI (root), `agent-runner`, and `harmonic-bridge`, with a note that these run natively on Node 22 with no containers and don't need the upgraded box (just disk for `node_modules`). Mirrors CI's `typescript` job exactly.
- **End-to-end tests (Playwright)** — `playwright:install` + `test:e2e`, flagged as the one TS surface that needs a running app stack, and that CI doesn't run it.

danallison:
let's put the readme changes in its own file under `docs/` instead of in the readme itself, and then have the readme point to that file. also, it should not mention the "upgraded box" which won't make any sense to future readers.

PR #315, opened 2026-06-30
Tier 1 Terraform provisioning + file-mounted secrets (draft skeleton)

melody-harmonic:
Draft for @danallison per [note 1cc7cc50](https://www.harmonic.social/collectives/harmonic-devs/n/1cc7cc50) — "proceed". This is the **skeleton to react to**, not a finished, applied tool.

## Why
The **release** path is already reproducible (CI → `deploy.sh` pull/up, rollback, maintenance mode). The gap is **provisioning + first-boot config**, which today is manual prose in `DEPLOYMENT.md`. That's Terraform's sweet spot — and the place where the secrets-in-state question has to be answered first.

## What's here
**Terraform — Tier 1, provision-only** (`terraform/`)
- Droplet (Docker via cloud-init), managed Postgres, Spaces bucket + scoped key, DNS incl. **wildcard** `*.<domain>` for tenant subdomains, firewall (only 80/443 public; `:8080`/metrics never exposed), AWS SES identity + DKIM/SPF written back into DO DNS.
- Release stays with `deploy.sh` — Terraform can't orchestrate drain/migrate/maintenance mode, so it shouldn't try.

**Secrets — the two axes**
- *Injection* (`docker-compose.secrets.yml` overlay + `config/initializers/00_file_secrets.rb`): move off leak-prone env vars to `/run/secrets` file mounts. Strict **no-op** until opted in (env still wins; file is fallback). Preserves agent-runner's existing minimal blast radius.
- *Storage* (`secrets/` + `deploy.sh`): **SOPS + age** encrypts a repo-committed secrets file; `deploy.sh` gains a **guarded** decrypt step that skips entirely when the encrypted file is absent. App-internal secrets (`SECRET_KEY_BASE`, `AGENT_RUNNER_SECRET`) → Rails credentials; infra/third-party → SOPS file.

**State**: only managed-resource creds (DB/Spaces, generated by the provider) unavoidably land in TF state → encrypted remote backend templated in `versions.tf`. Everything else stays out of state via SOPS.

Full rationale + the end-to-end flow: `docs/INFRASTRUCTURE.md`.

## Validation status
Statically validated locally (2026-07-03): `terraform fmt` + `terraform validate` (v1.9, real provider schemas), shellcheck on deploy/rollback/adapter scripts, compose prod+overlay merge check, `ruby -c`, `check-secrets.sh`. **Not yet `terraform apply`-ed against a live account.** Backward-compatible by construction: the initializer and the deploy.sh populate step are strict no-ops on today's hosts.

## Open questions (need your call)
1. **age vs AWS KMS** for the SOPS key — defaulted to age.
2. **Per-tenant secrets?** Layout assumes global secrets; per-tenant Stripe/etc. would reshape it. *Most worth confirming.*
3. **Secrets service (Infisical/SSM) now vs file-based-then-upgrade** — defaulted to file-based.


melody-harmonic:
Reworked the secrets design per Dan's directive that **no secret material — plaintext or ciphertext — may live in this open-source repo**, and to apply the two resolved votes.

**What changed (commit `f7cbb01`):**
- **Interface, not store.** The repo now ships only the contract: `secrets/secrets.example` (names, no values), the `/run/secrets/<NAME>` file-mount + Rails initializer (unchanged), and a `secrets/README.md`.
- **Swappable guarded hook.** `deploy.sh`'s hardcoded SOPS `decrypt_secrets()` → `populate_secrets()`: runs an optional `$SECRETS_HOOK` (default `/opt/harmonic/secrets/populate-secrets.sh`, **not shipped**) and enables the file-secrets overlay only when populated. Default is a strict no-op — operators pre-place files.
- **SOPS demoted to one optional adapter**, under `secrets/adapters/` (example hook + `.sops.yaml` template), explicitly sourcing the encrypted blob from a **private store outside this repo**. Dropped the "commit the cipher" example and the `.gitignore` line that permitted committing `*.enc.env`.
- **Compose overlay** reads from operator-populated `secrets/run/` (was `secrets/decrypted/`).
- **Docs:** `INFRASTRUCTURE.md` now leads with the no-secrets-in-repo constraint, frames adapters as operator choices, and records the resolved decisions.

**Applied votes:** secrets scope **global** ([bcf853b3](https://www.harmonic.social/collectives/harmonic-devs/d/bcf853b3), per-tenant got zero) → one shared set, no per-tenant path/split; topology **single-node** ([651b8093](https://www.harmonic.social/collectives/harmonic-devs/d/651b8093)).

Still **unvalidated** — no terraform/sops/test runner locally; repo's own `check-secrets.sh` passes clean. Vote e01c401d (SOPS+age vs KMS) is now effectively moot under the interface model — happy to re-cut it to "which adapters do we document" or close it, your call.

melody-harmonic:
Review pass per Dan's ask (commits `530a6593` merge of main + `ec9cb5b2`). Now that the box has Docker I could actually validate the pieces, and the review turned up real bugs in the secrets overlay:

**Bugs found & fixed**

1. **The overlay promised secrets it couldn't deliver.** `SMTP_*`/`REDIS_URL` are read in `config/environments/production.rb` and `POSTGRES_*` in `database.yml` ERB — all *before* `config/initializers/*` run, so `00_file_secrets.rb` fires too late for them. And production never reads `DATABASE_URL` at all (it uses split `POSTGRES_*` vars), so the overlay's marquee entry was a no-op. Separately, `REDIS_PASSWORD` and agent-runner's `AGENT_RUNNER_SECRET` are compose-interpolated (`${VAR:?}`) from *host* env — a container file mount can't satisfy that, so the whole agent-runner block was inert (Node reads `process.env` anyway). The overlay now mounts only what's genuinely read late: Spaces secret key, Stripe keys, GitHub OAuth secret, Turnstile key (newly added — it qualifies), and Rails-side `AGENT_RUNNER_SECRET`. The early-boot set stays on `.env` until we add a container entrypoint export (documented as the follow-up mechanism).
2. **`rollback.sh` stripped the overlay.** It recreated the stack with only the base compose file, so a rollback would remove `/run/secrets` mounts mid-incident. It now applies the same guard as `deploy.sh`.
3. **Partial migration would break `up`.** Verified empirically: compose errors (`bind source path does not exist`) if any declared secret file is missing. Documented convention: create the full file set, leave a file *empty* to keep that secret on `.env` (the initializer already ignores empty files).
4. **cloud-init contradicted the interface model** (still installed `age` + planted an `age.key` placeholder) and never copied `docker-compose.secrets.yml` to `/opt/harmonic`, so `deploy.sh` would fail on a fresh droplet once secrets were populated. Now adapter-neutral: 0700 `secrets/` + `secrets/run/` dirs, no key material in user-data, overlay file shipped.
5. **Terraform outputs aligned with reality:** `database_url` → `postgres_host/port/db/user/password`, matching what `database.yml` consumes.

**Validated** (previously the PR was honest-but-unvalidated): `terraform fmt` + `terraform validate` v1.9 against real provider schemas (so `digitalocean_spaces_key`, `graceful_shutdown`, etc. are schema-checked), shellcheck on deploy/rollback/all three adapter hooks, prod+overlay compose merge check, `ruby -c`, and `check-secrets.sh`. Also merged current `main` into the branch (clean). Still **not** applied to a live account — `terraform plan` against real creds is the remaining gate.

PR #317, opened 2026-06-30
Collective member management UI (roles + removal)

### LINKED ISSUE #316: collective members management ui (membeship status + roles)
currently there's no way to add or remove roles to collective members, and no way to remove a member from a collective after they've joined. we need a ui for admins to do that.

melody-harmonic:
Closes #316.

## Problem
Admins had no way to add/remove roles on collective members, and no way to remove a member once they'd joined.

## What this adds
Admin-only controls on the collective **Members** page (`/collectives/:handle/members`):
- **Role toggles** for `admin`, `representative`, and `summarizer` per member.
- **Remove** button per member (archives the membership — soft delete, consistent with `CollectiveMember#archive!`).

Controls render only for admins, and only on standard collectives (not the main collective or private workspaces). Non-admins see the existing read-only badge view unchanged.

## Endpoints
Mirroring the existing `add_ai_agent` / `remove_ai_agent` JSON pattern:
- `POST   /collectives/:handle/members/update_roles` → `CollectivesController#update_member_roles`
- `DELETE /collectives/:handle/members/remove` → `CollectivesController#remove_member`

## Guard rails
- Cannot revoke the `admin` role from the **last** admin (lockout protection).
- Cannot remove the **collective owner** (`created_by`).
- Admins cannot remove **themselves** via this UI (Leave handles that).
- Invalid roles → 422; non-members → 404; non-admins → 403.

## Implementation
- `app/javascript/controllers/collective_member_manager_controller.ts` — Stimulus controller (`collective-member-manager`) driving inline DOM updates, using `fetchWithCsrf`.
- View, CSS (Pulse tokens only), routes, controller actions + private helpers (`authorize_member_management`, `last_admin?`).

## Tests
- Controller tests: grant/revoke role, last-admin lockout, second-admin revoke, member removal, owner/self/non-admin/404 rejections, and admin-vs-non-admin rendering of the controls.
- Frontend Vitest: toggle (grant + active-state + error path) and remove (confirm + DELETE + card removal).

⚠️ **Not run locally** — this environment has no Docker, so the Rails/Vitest/Sorbet/RuboCop suites couldn't run here. Relying on CI to verify.


danallison:
the way it's designed currently, it looks like all members already have all roles, which is not what we want. instead, admins should see a kebab button next to each user with a dropdown list of buttons that follow "Add role [role]" or "Remove role [role]" depending on whatever roles the user currently has or doesn't have, and then a "Remove from collective" button at the bottom of the list.

app/services/actions_helper.rb
```diff
@@ -160,6 +160,26 @@ class ActionsHelper
       authorization: :collective_member,
       visibility: :by_collective,
     },
+    "update_member_roles" => {
+      description: "Grant or revoke a role on a member of this collective",
+      params_string: "(user_id, role, grant)",
+      params: [
+        { name: "user_id", type: "integer", required: true, description: "ID of the member to update" },
```
danallison: please use user handle instead of user id.

melody-harmonic: Done in c73808e4. Switched both `update_member_roles` and `remove_member` from `user_id` to a `user_handle` param (resolved tenant-scoped via TenantUser, leading `@` tolerated), and updated the members-page forms + controller tests to match. Changed `remove_member` too for consistency since it's the sibling action — let me know if you'd rather I'd left that one.

PR #342, opened 2026-07-01
Auto-hiding header that reveals on scroll-up (#338)

### LINKED ISSUE #338: auto-hiding header (reveals on scroll-up)

melody-harmonic:
Closes #338.

## What

The top app header now auto-hides on scroll-down and reveals on scroll-up (or when you return near the top of the page) — reclaiming vertical space while reading, especially on mobile.

## How

- **CSS** (`pulse/_layout.css`): `.pulse-top-header` is now `position: sticky; top: 0` with a `transform` transition. A new `.pulse-top-header--hidden` class translates it out of view (`translateY(-100%)`). `prefers-reduced-motion` drops the transition.
- **CSS** (`pulse/_responsive.css`): the mobile override keeps the header `sticky` (was `relative`) so it still auto-hides on mobile while remaining the containing block for the expanded search overlay.
- **JS**: new `auto-hide-header` Stimulus controller toggles the hidden class based on scroll direction, coalesced through `requestAnimationFrame`. It:
  - always reveals within the header's own height of the top,
  - ignores sub-threshold jitter (default 8px),
  - reveals when a descendant receives focus (keyboard a11y),
  - resets to visible on `connect` (guards against a stale Turbo-cached hidden state),
  - clamps negative iOS rubber-band overscroll offsets.
- Wired into the layout header and registered in `controllers/index.ts`.

The existing top-right dropdown menu positions itself relative to its `offsetParent`, which becomes the now-positioned header — so it continues to anchor correctly.

## Tests

Added `auto_hide_header_controller.test.ts` (vitest) covering hide-on-down, reveal-on-up, always-reveal-near-top, sub-threshold no-op, stale-snapshot reset, and focus reveal.

## Verification

Not verified locally — Docker isn't available on this VM and the disk was full so `npm install` couldn't run. Relying on CI for the frontend test suite + typecheck. Reviewed statically against existing controller conventions (Stimulus `classes`/`values` API, `Controller<HTMLElement>` generic) and the style-guide check (`pulse-` prefixed class, no hardcoded colors).

### DIFF
--- a/app/assets/stylesheets/pulse/_layout.css
+++ b/app/assets/stylesheets/pulse/_layout.css
@@ -67,6 +67,23 @@
   background: var(--color-canvas-default);
   min-height: 56px;
   box-sizing: border-box;
+  /* Auto-hiding header: pinned to the top, translated out of view on
+   * scroll-down and revealed on scroll-up by the auto-hide-header controller. */
+  position: sticky;
+  top: 0;
+  z-index: 200;
+  transition: transform 0.25s ease;
+  will-change: transform;
+}
+
+.pulse-top-header--hidden {
+  transform: translateY(-100%);
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .pulse-top-header {
+    transition: none;
+  }
 }
 
 .pulse-top-header .pulse-logo {
--- a/app/assets/stylesheets/pulse/_responsive.css
+++ b/app/assets/stylesheets/pulse/_responsive.css
@@ -30,7 +30,10 @@
 
   .pulse-top-header {
     padding: 8px 12px;
-    position: relative;
+    /* Stay sticky (not static) so the header still auto-hides on mobile and
+     * remains the containing block for the expanded search overlay below. */
+    position: sticky;
+    top: 0;
   }
 
   .header-search-wrapper {
--- a/app/javascript/controllers/auto_hide_header_controller.test.ts
+++ b/app/javascript/controllers/auto_hide_header_controller.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest"
+import { Application } from "@hotwired/stimulus"
+import AutoHideHeaderController from "./auto_hide_header_controller"
+
+const HIDDEN_CLASS = "pulse-top-header--hidden"
+
+describe("AutoHideHeaderController", () => {
+  let application: Application
+  let header: HTMLElement
+
+  // The controller coalesces scroll handling through requestAnimationFrame;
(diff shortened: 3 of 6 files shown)


melody-harmonic:
## Self-review of #342

I re-read the diff against the surrounding layout/CSS (scroll container, z-index stacking, dropdown/overlay anchoring). Overall this is solid — clean Stimulus controller, good test coverage, CI green. A few notes, none blocking:

**Verified correct**
- **Scroll container** — the controller listens on `window`. The app scrolls on the document (`body { display:flex; flex-direction:column; min-height:100vh }`; no internal `overflow:auto` scroll pane in the pulse layout, including chat mode where `.pulse-main` only zeroes padding). So window-scroll is the right target *today*.
- **z-index: 200** sits above page content (1–100) and below modals/toasts (1000 / 9999). The user menu (`.top-menu`, z-index 100) and mobile search overlay (z-index 100) are header *descendants*, so they render above the header regardless. Correct.
- **Dropdown anchoring** — making the header `sticky` makes it the `offsetParent` for `.top-menu` (`top:48px; right:16px`), which lands it just below the header just as before. The mobile `sticky` (was `relative`) keeps it a containing block for the expanded search overlay. Reasoning holds.

**Worth calling out**
1. **Behavior change beyond auto-hide:** the header was previously static/relative and scrolled away entirely; as `sticky` it now overlays content (and the sidebar top) scrolling *under* it while revealed. That's the intended tradeoff for reveal-on-scroll-up, but it's a real visual change — reviewers should confirm they want content passing under a ~56px bar mid-page.
2. **Load-bearing assumption:** if a future layout ever gives `.pulse-main` its own `overflow:auto` scroll pane, the header would silently stop auto-hiding there (window never scrolls). Worth a comment near the scroll listener.

**Minor / optional**
3. **Open dropdown + scroll-down:** if the user menu or expanded search is open and a scroll-down fires, the header (and its child overlay) translates off-screen. Low likelihood, self-corrects on scroll-up, and the `focusin` reveal partly covers the search case — but could suppress `hide()` while a menu/overlay is open.
4. **Style nit:** `update()` is `public` while its siblings (`hide`/`reveal`/`requestUpdate`) are `private`; nothing calls it externally (tests drive via scroll events), so it can be `private` for consistency.
5. **Test gap:** `disconnect()` listener cleanup isn't covered. Given Turbo connects/disconnects controllers often, a "removes scroll listener on disconnect" case would round out the suite. Optional.

None of these block merge. My recommendation: (1) and (2) each deserve a one-line code comment; the rest are optional polish. Happy to push the comments + the `update()` visibility fix if you want.

PR #348, opened 2026-07-02
Fix #335: primary button color for unconfirmed comment read-confirm

### LINKED ISSUE #335: Confirm read count icon color should match button color
If the current user has not confirmed reading a given comment, the clickable confirmed read count icon should appear as the primary button color, same as the button on feed item notes.

melody-harmonic:
Fixes #335.

## Problem
On a comment the current user has **not** confirmed reading, the clickable read-confirm count (`.pulse-comment-confirm-btn`) was styled with the muted foreground color and a transparent background — visually indistinguishable from static metadata, so it didn't read as an actionable control. The issue asks for it to appear in the primary button color, like the Confirm Read button on feed item notes.

## Fix
CSS only, in `pulse/_components.css`:
- The unconfirmed clickable button (`.pulse-comment-confirm-btn:not(.is-confirmed)`) now gets the filled primary treatment: `--color-fg-default` background, `--color-canvas-default` text — the same fill as `.pulse-action-btn` (the feed-item Confirm Read button). Hover inverts, mirroring `.pulse-action-btn:hover`.
- The shared base rule's `border: none` became `border: 1px solid transparent` so the primary variant can carry a visible border without changing box size relative to the confirmed / logged-out states (no cross-state layout shift within a thread).

Unchanged: the confirmed state (`.is-confirmed`, green) and the logged-out display span stay as they were.

## Verification
CSS-only change; no unit coverage for stylesheet rules. **No Docker on this VM**, so I could not render it locally — leaning on CI. Visual check recommended in light and dark mode (both use `--color-*` tokens, so both track automatically).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Unconfirmed comment read-confirm control uses the primary button color** (fixes #335) — the clickable read-confirm count on a comment you haven't confirmed now shows as a filled primary button, matching the Confirm Read button on feed item notes, instead of a muted icon.
+
 ## [1.37.0] - 2026-07-02
 
 ### Added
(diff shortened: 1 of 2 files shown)


danallison:
please remove the changelog entry. changelog gets updated when we bump the version.

melody-harmonic:
Done — removed the CHANGELOG entry and pushed. Understood: changelog gets updated at version-bump time, not per-PR. (Button-color fix itself unchanged.)

PR #349, opened 2026-07-02
Fix #344: settings page primary buttons render styled

### LINKED ISSUE #344: css bug: buttons on user settings page
many of the buttons on the user settings page do not use the correct button styles. there might be other pages also.

melody-harmonic:
Fixes #344.

## Root cause
`pulse-action-btn-primary` is a **modifier** — there is no standalone `.pulse-action-btn-primary` rule in the stylesheets (grep confirms zero matches). It's meant to pair with the base `.pulse-action-btn`, which is how ~53 other usages apply it. Four buttons applied it **alone**, so they matched no button rule and fell back to unstyled native browser buttons:

| Button | Location |
|---|---|
| Save Changes (profile) | `users/settings.html.erb` |
| Save (workspace Trio) | `users/settings.html.erb` |
| Change Email | `users/settings.html.erb` |
| Save notification preferences | `shared/_notification_preferences.html.erb` (rendered on the settings page; also reusable elsewhere) |

That matches the report: "many of the buttons on the user settings page do not use the correct button styles… there might be other pages also."

## Fix
Add the base `pulse-action-btn` class to all four. View-only; no behavior change.

Note: `.pulse-action-btn-primary` still has no distinct visual treatment of its own (base `.pulse-action-btn` is already the filled/dark style), so "primary" currently reads identically to a default action button across the whole app. Whether primary buttons should get a distinct accent is a design call I left for you rather than fold into this bug fix — happy to open a follow-up if you want it.

## Verification
View-only. **No Docker on this VM**, so no local render — leaning on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-### Fixed
-
-- **Settings page primary buttons render styled** (fixes #344) — "Save Changes", "Change Email", the workspace-Trio "Save", and "Save notification preferences" used `pulse-action-btn-primary` on its own. That class is only a modifier (no standalone rule), so those buttons fell back to unstyled browser defaults. Paired them with the base `pulse-action-btn`.
-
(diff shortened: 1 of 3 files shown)


danallison:
please remove the changelog entry. changelog gets updated when the version gets bumped.

melody-harmonic:
Done — removed the CHANGELOG entry and pushed. (Settings-button style fix itself unchanged.)

PR #350, opened 2026-07-02
Fix #346: current device shows "Active now" in device list

### LINKED ISSUE #346: bug: devices list in settings has incorrect "last used at"
the current device last used at shows things like "7 hours ago" which is obviously wrong since the device is currently in use.

melody-harmonic:
Fixes #346.

## Root cause
A `RefreshToken`'s `last_used_at` is only written when the token **rotates** — i.e. on a silent re-auth (`RefreshToken#rotate!`) or at initial issuance. Rotation happens only after the session cookie's timestamps go stale and the controller mints a new session. While your session is live, no rotation occurs, so `last_used_at` sits at the time of your last login/refresh. On the **current** device that produces the reported "Last used 7 hours ago" while the device is obviously in use.

## Fix
The current device is, by definition, being used right now (it's the one making the request — `is_current` already computed from `current_refresh_token`). Show **"Active now"** for it instead of the stale relative time. Other devices keep `Last used … ago`, which remains the best available estimate — we can't know a remote device's live activity without it hitting the server, and rotation is the only server-side "used" signal we get.

## Scope note (the "trivial vs small" question)
This is the **trivial** view-only fix and it fully resolves the reported symptom (current device). The deeper "make `last_used_at` reflect real recent activity for *all* devices" would mean touching the authenticated request path to bump the timestamp (throttled, to avoid a write per request) — a bigger change with its own tradeoffs. I did not do that here; happy to open a follow-up if you want live last-used for non-current devices.

## Verification
View-only. **No Docker on this VM**, so no local render — leaning on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Current device shows "Active now" in the device list** (fixes #346) — a device's "last used" only advances when its refresh token rotates (a silent re-auth), so the device you're currently on showed a stale time ("7 hours ago") while plainly in use. The current device now reads "Active now"; other devices keep the last-used estimate.
+
 ## [1.37.0] - 2026-07-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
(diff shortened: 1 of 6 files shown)


danallison:
this is not the right fix. please fix the issue correctly so that all devices display the correct timestamp. and also remove the changelog entry. changelog gets updated at version bump time.

melody-harmonic:
You're right, the "Active now" display was a band-aid. Reworked it to fix the data instead:

- Root cause: a refresh token's `last_used_at` only advanced on rotation (silent re-auth), and a live session never rotates — so the device you're actively on stayed stale.
- Now ordinary request activity refreshes the current device's row via `RefreshToken#touch_last_used!`, called from a new `touch_current_device_activity` before_action in `ApplicationController`. Throttled (`ACTIVITY_TOUCH_THROTTLE = 5.minutes`) so it's at most one `update_column` per window, not a write per request. No-op on revoked/expired tokens, and it only touches the device whose cookie made the request — so every device shows its own accurate time.
- The view now shows `last_used_at` uniformly for all devices (dropped the `is_current` special-case).
- Removed the CHANGELOG entry.

Added model tests for the throttle/revoked/expired paths and an integration test that an ordinary request refreshes the current device (and leaves others alone). Usual caveat: no Docker on this VM, so leaning on CI to run them.

PR #366, opened 2026-07-03
Fix collective representation over markdown/MCP (#365)

### LINKED ISSUE #365: Collective representation can't be started via markdown/MCP (start_representation action + represent.md.erb missing)
## Summary

An agent that holds the **representative** role in a collective cannot start a **collective representation session** through the markdown/MCP interface. Both discovery paths a markdown client relies on are broken for collective representation:

- `fetch_page /collectives/:handle/represent` → **HTTP 406** (no markdown template), so the client can't read the page or its action list.
- `execute_action start_representation` → **"No actions are defined at this path"** at every plausible collective path (`/represent`, `/representation`, collective root — root only exposes `send_heartbeat`).

Net effect: the representative role can be granted, but a markdown/MCP agent has no reachable way to *use* it for a collective. This surfaced when @melody-harmonic was asked to post a public announcement as `harmonic-devs` and had the role but could not open a session.

Note this is specific to **collective** representation. **Trustee-grant (user) representation works** via dedicated describe/execute action routes, which is exactly the pattern the collective path is missing.

## Root causes

**1. Missing markdown template → 406**
`RepresentationSessionsController#represent` responds to both `html` and `md` (`app/controllers/representation_sessions_controller.rb`), but only `app/views/representation_sessions/represent.html.erb` exists — there is no `represent.md.erb`. So an `Accept: text/markdown` request has no template and Rails returns 406.

```
$ ls app/views/representation_sessions/
index.html.erb  index.md.erb  represent.html.erb  representing.html.erb
representing.md.erb  show.html.erb  show.md.erb   ...
# represent.md.erb is absent
```

**2. No MCP action surfaced for collective representation**
The collective represent routes (`config/routes.rb:577-584`) only expose the HTML form flow:
```ruby
get  "#{prefix}/represent" => 'representation_sessions#represent'          # form page (no .md)
post "#{prefix}/represent" => 'representation_sessions#start_representing' # form submit (needs understand=true)
```
There is **no `describe`/`execute` action pair** for it, and no entry in `app/services/actions_helper.rb` (`ACTIONS`) for the `/collectives/:collective_handle/represent` path — so `available_actions_for_current_route` never lists a `start_representation` action for a collective, and `execute_action` finds nothing.

Contrast with **trustee-grant (user) representation**, which *does* work over MCP because it has the action routes (`config/routes.rb:458-461`):
```ruby
get  'settings/trustee-authorizations/:grant_id/actions/start_representation' => 'trustee_grants#describe_start_representation', on: :member
post 'settings/trustee-authorizations/:grant_id/actions/start_representation' => 'trustee_grants#execute_start_representation', on: :member
# ...and matching end_representation pair
```

**3. Help doc doesn't name the path**
`app/views/help/agents/representation.md.erb` tells agents to "call `execute_action` with the `start_representation` action" and states a collective can be represented with just the representative role, but never names the path the action lives on or the params — so even a correct reading of the doc dead-ends for collectives.

## Suggested fix

Give collective representation the same describe/execute action surface as trustee grants, so `execute_action start_representation` works on the collective:

- Add `.../represent/actions/start_representation` (and `end_representation`) describe+execute routes → `RepresentationSessionsController`, mirroring `trustee_grants#{describe,execute}_start_representation`.
- Register the action(s) in `ACTIONS` (`app/services/actions_helper.rb`) under the collective path, gated on `can_represent?` so they only appear to eligible members.
- Either add `represent.md.erb` or drop `format.md` from `#represent` so the page 404s cleanly instead of 406-ing. (Lower priority once the action exists, since the action is what MCP needs.)
- Update `help/agents/representation.md.erb` to name the path and params for collective representation.

## Repro

1. As an agent with the representative role in a collective, `fetch_page /collectives/<handle>/represent` → 406.
2. `execute_action start_representation` on the collective / `/represent` / `/representation` → "No actions are defined at this path".

## Environment
Observed on www.harmonic.social via the MCP markdown interface, 2026-07-03. Reported by @melody-harmonic; @danallison to handle the public announcement in the meantime.

melody-harmonic:
Fixes #365.

## Problem

An agent that holds the **representative** role in a collective couldn't start a **collective** representation session through the markdown/MCP interface. Both discovery paths were broken:

- `GET /collectives/:handle/represent` → **406** (no markdown template).
- `execute_action start_representation` → **"No actions are defined at this path"** — there was no describe/execute action pair for the collective, only the HTML form flow.

This is specific to *collective* representation; **trustee (user) representation already works** because it has the pieces the collective path was missing — that's the fix template used here.

## Changes

- **Routes** (`config/routes.rb`): add `/collectives/:handle/represent/actions/{start,end}_representation` (describe `GET` + execute `POST`) and an `/actions` index, mirroring the trustee-grant routes.
- **Controller** (`RepresentationSessionsController`): add `describe`/`execute` methods for `start_representation`/`end_representation` and an `actions_index_represent`, mirroring `TrusteeGrantsController`. `#represent` now also responds to `format.md`.
- **ActionsHelper**: register the represent route with conditional `start_representation`/`end_representation` actions, gated on `can_represent?` and mutual exclusivity, so markdown-frontmatter discovery agrees with execute-time enforcement.
- **View**: add `represent.md.erb` so the page renders instead of 406-ing.
- **Help doc**: name the collective path + params in `help/agents/representation.md.erb`.

Reuses the existing `start_representation`/`end_representation` `ACTION_DEFINITIONS` and `CapabilityCheck` entries — no visibility/capability-list changes.

## Enforcement

- Discovery: `start_representation` shows only to a member who `can_represent?` with no active session; `end_representation` shows once a session is open.
- Execution: `execute_start_representation` re-checks `can_represent?`, archived state, and nested-session conflict; `execute_end_representation` resolves the session by `(collective, representative)` and requires the caller to be its representative. Capability layer still gates both via the existing grantable `start_representation`/`end_representation` capabilities.

## Tests

Adds `test/integration/api_collective_representation_test.rb` covering: represent page renders markdown and advertises the action, start (happy path + role enforcement + nested-session rejection), and end (happy path + no-active-session).

Run locally via docker-compose.dev.yml: all 11 tests in this file pass, plus the 64 tests in the neighboring representation suites (api_representation, representation, representation_session).

### DIFF
--- a/app/controllers/representation_sessions_controller.rb
+++ b/app/controllers/representation_sessions_controller.rb
@@ -56,6 +56,16 @@ class RepresentationSessionsController < ApplicationController
     end
 
     @page_title = "Represent"
+    # DB-backed lookup mirroring the frontmatter lambdas in ActionsHelper: an API
+    # caller isn't required to echo X-Representation-Session-ID on this page, so
+    # current_representation_session alone would say "no session" while the
+    # frontmatter advertises end_representation. The prose must agree with the
+    # advertised actions.
+    @active_collective_session = RepresentationSession.find_by(
+      collective: current_collective,
+      representative_user: @current_user,
+      ended_at: nil
+    )
     respond_to do |format|
       format.html
       format.md
--- a/app/controllers/representation_sessions_controller.rb
+++ b/app/controllers/representation_sessions_controller.rb
@@ -8,7 +8,12 @@ class RepresentationSessionsController < ApplicationController
   include RequiresReverification
 
   before_action :set_sidebar_mode, only: [:index, :show, :represent, :representing]
-  before_action -> { require_reverification(scope: "representation") }, only: [:represent, :start_representing, :start_representing_user]
+  # execute_start_representation is included so a browser/cookie caller hitting the
+  # actions endpoint faces the same 2FA step-up as the HTML form flow. API-token
+  # callers are exempt inside require_reverification (returns early), so MCP is
+  # unaffected — but this closes the browser bypass. End is intentionally not gated
+  # on either path.
+  before_action -> { require_reverification(scope: "representation") }, only: [:represent, :start_representing, :start_representing_user, :execute_start_representation]
 
   def index
     @representatives = current_collective.representatives
@@ -83,7 +88,15 @@ class RepresentationSessionsController < ApplicationController
   end
 
   def execute_start_representation
-    if current_representation_session
+    # Nested-session guard. `current_representation_session` resolves from the
+    # X-Representation-Session-ID header for API callers, and nothing forces that
+    # header on start — so an agent could start, then start again without echoing
+    # the header and slip past a header-only check, creating a second concurrent
+    # active session (there's no partial-unique DB index as a backstop). Guard on
+    # the database instead so a live session for this (collective, representative)
+    # always blocks a second start regardless of headers.
+    if current_representation_session ||
+       RepresentationSession.exists?(collective: current_collective, representative_user: current_user, ended_at: nil)
       return render_action_error({
                                    action_name: "start_representation",
                                    resource: current_collective,
@@ -97,6 +110,20 @@ class RepresentationSessionsController < ApplicationController
                                    action_name: "start_representation",
                                    resource: current_collective,
                                    error: "You do not have permission to represent this collective.",
+                                   status: :forbidden,
+                                 })
+    end
+
+    # Don't let an agent start a session it can't end. end_representation routes
+    # through the capability layer (unlike the HTML stop path, which is exempt via
+    # SESSION_MANAGEMENT_WRITES), so an agent configured with start_representation
+    # but not end_representation would start a collective session and then be
+    # denied ending it over MCP — stuck until the session expires. Refuse up front.
+    if CapabilityCheck.missing_rep_lifecycle_capabilities(@current_user).include?("end_representation")
+      return render_action_error({
+                                   action_name: "start_representation",
+                                   resource: current_collective,
+                                   error: "Your agent configuration lacks the end_representation capability, so you would be unable to end this session. Grant end_representation before starting.",
                                    status: :forbidden,
                                  })
     end
--- a/app/controllers/representation_sessions_controller.rb
+++ b/app/controllers/representation_sessions_controller.rb
@@ -41,13 +41,137 @@ class RepresentationSessionsController < ApplicationController
     # User representation is initiated from the trustee authorizations settings page instead.
     @can_represent_collective = @current_user.collective_member&.can_represent?
 
-    if @can_represent_collective
-      @page_title = "Represent"
-    else
+    unless @can_represent_collective
       # TODO: - design a better solution for this
       @sidebar_mode = "minimal"
-      render layout: "application", html: "You do not have permission to access this page."
+      return respond_to do |format|
+        format.html { render layout: "application", html: "You do not have permission to access this page." }
+        format.md { render plain: "# Represent\n\nYou do not have permission to represent this collective.", status: :forbidden }
+      end
     end
+
+    @page_title = "Represent"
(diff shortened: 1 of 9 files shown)


melody-harmonic:
Self-review pass (requested by Dan). Verified against the trustee-grant implementation it mirrors:

- Routes/controller/ActionsHelper wiring matches the trustee describe/execute pattern; the conditional-action lambdas use the same context keys (`collective`, `user`) that `build_condition_context` provides, so fetch_page frontmatter and the `/actions` index agree.
- `require_reverification` confirmed to exempt API-token callers (`return if api_token_present?`), so the added `execute_start_representation` gate only affects browser/cookie callers, as intended.
- The DB-backed nested-session guard and the end-capability preflight both hold up; existing `ACTION_DEFINITIONS` / `CapabilityCheck` entries are reused, so no visibility-list changes needed.

One real issue found and fixed (acdf996): `represent.md.erb` branched on the header-derived `current_representation_session`, while the frontmatter lambdas are DB-backed. An agent that started a session and re-fetched the page **without** echoing `X-Representation-Session-ID` (nothing requires it there) was told it could *start* a session while the frontmatter correctly advertised only `end_representation`. The prose now uses the same DB lookup, surfaces the active session's ID, and points at `end_representation`. Added a regression test.

Tests now run locally via docker-compose.dev.yml: 11/11 in this file, 64/64 across the neighboring representation suites. PR body's stale "no Docker here" warning removed.

danallison:
representation sessions work a bit differently when using mcp and not direct http. mcp users cannot send http headers. please make sure you understand the full end-to-end flow from the perspective of an mcp client and write the markdown copy for mcp only. direct http is a special case and should not be assumed. also make sure to review the help pages and make sure everything is consistent with these changes.

melody-harmonic:
Understood — reworked in 87864ef. My understanding of the end-to-end flow from an MCP client's seat:

1. The agent calls `execute_action` with `start_representation` at `/collectives/:handle/represent` as itself (plain `context`, no representation fields). The response returns the session ID.
2. On every subsequent tool call made on the collective's behalf, the agent declares the session **inside the tool arguments**: `context.representation_session_id` plus `identity.acting_as` (writes) or `identity.viewer`/`identity.viewing_as` (reads), with the represented handle being the collective's handle. It never touches HTTP headers — the MCP endpoint extracts that pair (`representation_pair`) and `MarkdownUiService#with_representation` attaches the `X-Representation-Session-ID`/`X-Representing-*` trio to the internal dispatch itself.
3. Ending is `execute_action` with `end_representation` at the same path, again as itself.

Copy changes, all written for that flow only (no header or `DELETE /representing` instructions anywhere agent-facing):

- **`represent.md.erb`** — active-session branch now says to declare `context.representation_session_id` + `identity.acting_as` `@<collective-handle>`, links `/help/agents/representation`, and ends via the `end_representation` action; the pre-session branch points at the context block instead of a header.
- **`start_representation` success messages** — both the new collective path and the existing trustee path (`trustee_grants_controller.rb`) told the agent to use the header; both now give the context-block fields with the concrete handle to declare.
- **`/representing` page** — added a "Using this session" section with the same fields, and the "Ending the session" copy now points only at the `end_representation` action (was suggesting `DELETE /representing` + header).

Help-page consistency pass: `/help/agents/representation` was already MCP-only (context block, no headers) and agrees with the above, as do `/help/representation` and `/help/mcp`. The one inconsistency found was `/help/collectives`, which said to start a session at `{collective}/representation` — that's the session log; it now points to `{collective}/represent` and lists both URLs. Internal code comments still name the `X-Representation-*` headers since that layer genuinely is header-based; only agent-addressed copy changed. `docs/REPRESENTATION.md` (developer docs) still documents the direct-HTTP header flow — left as is, since that's the special case documented for developers rather than copy served to agents; happy to add an MCP note there too if you'd like.

Both representation test files pass locally (42 runs, 0 failures).

PR #389, opened 2026-07-04
Notify attendees when a calendar event is about to start

### LINKED ISSUE #313: Calendar event notifications

melody-harmonic:
Fixes #313

## Design

Follows the `DeadlineEventJob` pattern exactly:

- **`CalendarEventStartingJob`** (new, minutely via sidekiq-cron) scans all tenants for calendar events with `starts_at` within a 1-hour lead window, fires a `commitment.starting` event, and marks a new `commitments.starting_event_fired_at` column *before* dispatching so a mid-flight failure can't double-fire. An `ends_at > now` guard keeps the pre-existing backlog of past events from firing a notification blast on deploy.
- **`NotificationDispatcher`** handles `commitment.starting` by notifying the creator plus all committed participants ("Starting soon: <title>", body = time range · location, linking to the event).
- **New notification type `event_starting`** ("Event starting soon") added to `Notification::NOTIFICATION_TYPES`, `DEFAULT_NOTIFICATION_PREFERENCES` (in-app + web push on, email off), and `NOTIFICATION_TYPE_LABELS`, so users can tune channels per the existing preferences UI. The registry sync test covers this.
- **Reschedule handling**: a `before_save` hook on Commitment clears `starting_event_fired_at` when `starts_at` moves to a time outside the lead window — a rescheduled event re-notifies for its new time, but nudging an imminent event by a few minutes doesn't re-ping (pairs with #321/PR #385, which makes event dates editable).
- `commitment.starting` also flows to webhooks/automations like any other event; documented in the automation YAML reference.

## Tests

8 job tests mirroring `deadline_event_job_test.rb`: fires within lead window, notifies creator + participants, skips far-future / already-ended / soft-deleted events, no double-fire, re-arm on reschedule, no re-arm on imminent nudge. Ran alongside the deadline job, tenant-user registry, and cron-schedule suites (71 runs, 0 failures); `srb tc` clean (Commitment RBI regenerated).

Note: `db/structure.sql` here contains only this branch's change; if this merges after #388 (which also touches structure.sql) the schema_migrations insert may need a trivial conflict resolution.

### DIFF
--- a/app/jobs/calendar_event_starting_job.rb
+++ b/app/jobs/calendar_event_starting_job.rb
@@ -1,69 +1,111 @@
 # typed: true
 # frozen_string_literal: true
 
-# CalendarEventStartingJob fires a commitment.starting event shortly before a
-# calendar event's start time, so attendees get a heads-up notification (see
-# NotificationDispatcher#handle_commitment_starting_event). It runs every
-# minute via sidekiq-cron and mirrors DeadlineEventJob: query across all
-# tenants, mark fired before dispatching so a failure can't re-fire.
+# CalendarEventStartingJob fires the two start-related events for calendar
+# events, so attendees get notified both ahead of time and at the moment the
+# event begins (see NotificationDispatcher#handle_commitment_starting_soon_event
+# and #handle_commitment_starting_event):
+#
+#   - commitment.starting_soon — a heads-up LEAD_TIME before starts_at
+#   - commitment.starting      — when starts_at is reached
+#
+# It runs every minute via sidekiq-cron and mirrors DeadlineEventJob: query
+# across all tenants, mark fired before dispatching so a failure can't re-fire.
 class CalendarEventStartingJob < SystemJob
   extend T::Sig
 
   queue_as :default
 
   MAX_PER_RUN = 100
-  # How far before starts_at the heads-up fires.
+  # How far before starts_at the "starting soon" heads-up fires.
   LEAD_TIME = 1.hour
+  # How long after starts_at we'll still fire the "starting" event. Keeps the
+  # backlog of long-running events that were already underway at deploy time
+  # (or when the job was down) from firing a stale "starting now" blast; the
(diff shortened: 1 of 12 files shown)


danallison:
i think we need to be more clear and explicit about the difference between a calendar event _starting_ vs _starting soon_, and we should have notifications for both. ideally, users could configure how far in advance to be notified/reminded about a given event, but that's out of scope for now. for now, let's just make the two notification events: starting and starting soon.

melody-harmonic:
Done in 9ccfa5ca — split into two distinct events, each with its own notification type, fired-at column, and reschedule re-arm:

- **`commitment.starting_soon`** → `event_starting_soon` ("Starting soon: …"): the heads-up, fires `LEAD_TIME` (1h) before `starts_at`.
- **`commitment.starting`** → `event_starting` ("Starting now: …"): fires when `starts_at` is actually reached. Bounded by a `STARTED_GRACE` (5m) window so the backlog of already-underway events doesn't blast a stale "starting now" on deploy.

The single minutely job now processes both scopes. Per-event advance-notice timing is still out of scope as you noted. Registry/prefs/labels and the automation event reference cover both; 12 job tests green (soon + starting fire/notify/dedup/reschedule paths), `srb tc` clean.

Still `CONFLICTING` on `db/structure.sql` vs #388 — trivial column-list + schema_migrations resolution at merge, same as flagged in the description.

danallison:
please resolve the merge conflict

melody-harmonic:
Conflict resolved — rebased onto `origin/main` (which had advanced past #380). Only `db/structure.sql` collided: kept both migration versions (`20260704010000` from this branch alongside main's `20260703120000`), no schema divergence. Now **MERGEABLE**, CI re-running.

While here I also rebased the two sibling PRs that #380's merge had re-conflicted the same way — #388 (structure.sql) and #387 (a test-file overlap where main's calendar-schedule tests and this branch's critical-mass tests were inserted at the same spot; kept both). All three are green-pending and mergeable now.

Issue #383, opened 2026-07-04
bug: can't navigate public space when in representation session


danallison:
homepage is redirected

PR #417, opened 2026-07-06
Keep header nav add + profile controls visible while representing (#415)

### LINKED ISSUE #415: bug: header nav add button and profile button disappear when representing a collective

melody-harmonic:
Fixes #415.

## Problem
Starting a representation session made the header's **add (+) button and profile/avatar menu disappear**. Both the desktop header (`top_right_menu`) and the mobile bottom tab bar's **You** tab were wrapped in `unless @current_representation_session`, so representation stripped:

- the collective **+** add button (the exact control needed to create a note/decision as the collective),
- search + inbox, and
- the avatar/account menu (on mobile, the only route to it).

That's the wall between us and posting the release announcement as `@harmonic-devs`: you can start the session, but there's no compose affordance while it's active.

## Fix
Drop both `unless @current_representation_session` guards. The represented identity keeps its full header control surface. The `You are acting on behalf of …` session banner (with its own **End Session** button) already carries the representation context, so nothing is lost by keeping these visible.

`@current_user` is already the represented identity during a session (`effective_user`), so the avatar and `+` button naturally reflect the collective you're acting as.

## Tests
Added a regression test (`test/integration/representation_test.rb`) asserting the add button, avatar menu, and mobile You tab all render during a representation session. Verified it **fails on the pre-fix guards** (`Expected 0 to be >= 1` for the add button) and **passes with the fix**.

Ran locally (docker dev): `representation_test.rb` + `api_collective_representation_test.rb` — 29 runs, 101 assertions, 0 failures.

## Context
Third and final blocker in the chain to post the v1.39/v1.40 announcement as the collective, after #404 (act-under a collective session over MCP) and #405 (#383, public space navigable while representing).

### DIFF
--- a/app/components/places_sheet_component.html.erb
+++ b/app/components/places_sheet_component.html.erb
@@ -13,14 +13,16 @@
         <span class="pulse-places-badge pulse-places-row-badge" data-collective-id="<%= @main_collective.id %>"
               style="<%= count_badge_style(unread_count_for(@main_collective)) %>"><%= count_badge_text(unread_count_for(@main_collective)) %></span>
       <% end %>
-      <%= tag.a href: "/chat",
-                class: ["pulse-places-row", ("active" if chat_active?)],
-                aria: { current: ("page" if chat_active?) } do %>
-        <span class="pulse-places-indicator"></span>
-        <span class="pulse-places-row-icon"><%= helpers.octicon "comment-discussion", height: 20 %></span>
-        <span class="pulse-places-row-name">Chat</span>
-        <span class="pulse-places-badge pulse-places-row-badge" data-chat-badge
-              style="<%= count_badge_style(@chat_unread_count) %>"><%= count_badge_text(@chat_unread_count) %></span>
+      <% if @show_chat %>
+        <%= tag.a href: "/chat",
+                  class: ["pulse-places-row", ("active" if chat_active?)],
+                  aria: { current: ("page" if chat_active?) } do %>
+          <span class="pulse-places-indicator"></span>
+          <span class="pulse-places-row-icon"><%= helpers.octicon "comment-discussion", height: 20 %></span>
+          <span class="pulse-places-row-name">Chat</span>
+          <span class="pulse-places-badge pulse-places-row-badge" data-chat-badge
+                style="<%= count_badge_style(@chat_unread_count) %>"><%= count_badge_text(@chat_unread_count) %></span>
+        <% end %>
       <% end %>
       <div class="pulse-places-divider"></div>
     <% end %>
--- a/app/components/places_sheet_component.rb
+++ b/app/components/places_sheet_component.rb
@@ -17,16 +17,20 @@ class PlacesSheetComponent < ViewComponent::Base
       collectives: T::Array[Collective],
       current_path: T.nilable(String),
       unread_counts: T::Hash[String, Integer],
-      chat_unread_count: Integer
+      chat_unread_count: Integer,
+      show_chat: T::Boolean
     ).void
   end
-  def initialize(main_collective: nil, collectives: [], current_path: nil, unread_counts: {}, chat_unread_count: 0)
+  def initialize(main_collective: nil, collectives: [], current_path: nil, unread_counts: {}, chat_unread_count: 0, show_chat: true)
     super()
     @main_collective = main_collective
     @collectives = T.let(collectives.select { |c| c.path.present? }, T::Array[Collective])
     @current_path = current_path
     @unread_counts = unread_counts
     @chat_unread_count = chat_unread_count
+    # Chat is hidden while representing — the represented identity's DMs aren't
+    # the representative's to see, and /chat is blocked during representation.
+    @show_chat = show_chat
   end
 
   sig { returns(T::Boolean) }
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -829,6 +829,9 @@ class ApplicationController < ActionController::Base
   def places_chat_unread_count
     return @places_chat_unread_count if defined?(@places_chat_unread_count)
     return (@places_chat_unread_count = 0) if @current_user.nil? || current_tenant.nil?
+    # Chat is hidden while representing; keep its badge count at zero so the
+    # broadcast never resurrects a count for a row that isn't shown.
+    return (@places_chat_unread_count = 0) if @current_representation_session
 
     @places_chat_unread_count = NotificationService.unread_chat_count_for(@current_user, tenant: current_tenant)
   end
@@ -915,7 +918,13 @@ class ApplicationController < ActionController::Base
 
     # Load notification count for HTML and markdown UI, but not JSON API.
     @unread_notification_count = if @current_user && current_tenant && !request.format.json?
-                                   NotificationService.unread_count_for(@current_user, tenant: current_tenant)
+                                   count = NotificationService.unread_count_for(@current_user, tenant: current_tenant)
+                                   # While representing, chat_message rows are hidden from the
+                                   # inbox — keep the header badge total in step by excluding them.
+                                   if @current_representation_session
+                                     count -= NotificationService.unread_chat_count_for(@current_user, tenant: current_tenant)
+                                   end
+                                   count
                                  else
                                    0
                                  end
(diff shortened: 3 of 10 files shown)


danallison:
do we know what the original intention of the conditional was? i imagine there was a reason for the check. for example, this now reveals the "sign out" button, which doesn't really make sense in a representation session. it also reveals the link to the represented user's private workspace (which doesn't exist for collective identity users). there might be other reasons i haven't thought of.

also, for collective representation, it makes sense to show notifications. but for user representation we need to think carefully about what the right thing to do is. it might depend on which capabilities are authorized. notifications can reveal chat message content that would not otherwise be accessible to a representative.

melody-harmonic:
## CI failure: root cause

The failing job is `rails-tests (integration)`, and the failing test is **not** the one I added — it's a pre-existing guard:

```
PrivateWorkspaceTest#test_representing_page_does_not_list_the_represented_user's_private_workspace
  test/integration/private_workspace_test.rb:201
  "Representing page should not list Bob's private workspace"
```

It only shows up in CI (not when I run `representation_test.rb` alone) because it lives in a different file and the whole `test/integration` dir runs together. It reproduces locally with the CI seed. **This is a real regression, not flakiness** — and it's exactly the leak you flagged.

Mechanism: during representation `@current_user` is the *represented* identity (user rep → the represented person; collective rep → the collective's `identity_user`). By deleting the `unless @current_representation_session` guard I brought the full `user_menu` back for representation sessions, so during **user** representation the represented person's **private-workspace link** renders in the header dropdown — which the representative must not see. That test catches it.

## Answering your questions — what the guard was actually protecting

The original guard hid the whole `user_menu`, which is full of items that are wrong or leaky when the avatar is a *represented* identity:

- **Sign Out** (`logout_path`) — ends the underlying human login entirely. Wrong exit; End Session is the right one (and it's already in the banner).
- **Workspace link** (`@current_user.private_workspace`) — for collective rep the `identity_user` has none (nil-guarded, never shown); for **user** rep it's the represented user's private workspace → the leak above.
- **Settings / AI Agents / Subdomains / Admin / Lists** — all manage the human's own identity/tenant; confusing when acting-as.
- **Notifications inbox** — the represented identity's notifications. Collective rep: the collective's — fine, as you said. User rep: the represented person's — can expose chat/DM content the representative isn't entitled to. Capability-dependent, agreed.

So a blanket "always render `top_right_menu`" is too blunt.

## Proposed direction

#415 is specifically about **collective** representation losing the add (+) and profile buttons. I'd scope the fix to that and treat the menu contents deliberately:

1. **Keep visible during representation:** the add (+) button and search — these are about acting *as* the represented identity, which is the whole point.
2. **Replace `user_menu` during representation** with a minimal, representation-appropriate menu: **Profile** (of the represented identity) + **Help**. Drop Sign Out, Workspace, Settings, AI Agents, Subdomains, Admin. End Session stays in the banner — no need to duplicate it.
3. **Notifications:** show for **collective** representation; **hide for user representation** by default (safer), and revisit as a real capability once the grant model can express "may read notifications."

Two calls I'd like from you before I rewrite the commit:

- **D1 — Notifications during *user* representation:** hide by default (my recommendation), gate on a capability now, or show?
- **D2 — Menu contents during representation:** the minimal set (Profile + Help), or would you keep a few more items?

Happy to formalize D1/D2 as a vote if you'd prefer that over the thread.

danallison:
i worry we might end up in a whack-a-mole situation if we make a rule that says representatives can't see notifications. not only do notifications appear in the top nav (or bottom bar on mobile) and the notifications page, there's also notifications that appear in the places sheet and then there's all of the `my:` search filters, including `my:notified`. if we're going to hide chat messages specifically rather than just notifications, that's probably more doable because those messages only show up on the notifications page, the places sheet (unread message count), and `/chat` itself. it looks like navigating to `/chat` already ends the representation session, even for collective representation. i think chat message notifications are already grouped together on the notifications page, so hopefully it's not difficult to just hide that group if there's an active representation session. and it should be straightforward to hide the chat link in the places sheet. that should cover it i think.

melody-harmonic:
Took your direction — **hide chat specifically, not notifications** — and while implementing it I found the assumption that `/chat` already ends/blocks the session doesn't hold. Pushed `3c369f2`.

### ⚠️ `/chat` was not actually protected
I tested it directly: during a **user** representation session, `GET /chat` returned **200** (not a redirect), so the representative could reach the represented user's DMs. Same for `/notifications` (200). The nav guard in `current_representation_session` that looks like it restricts representation to `/collectives/*` never fires in the browser flow — `@current_representation_session` is already memoized by the time that getter runs, so it early-returns before the redirect. So chat needed an explicit block; it wasn't getting one.

### What I changed (chat-specific, per your whack-a-mole point)
- **`/chat` blocked during representation** — `ChatsController` before_action redirects browsers to `/representing`, 403 for agent/polling callers. Covers `/chat` and `/chat/:handle`. Collective rep loses nothing (no DMs).
- **Notifications page** drops `chat_message` rows while representing (they carry no event, so they land in the "Other" group mixed with reminders — I filter by type, not by hiding a group, so reminders stay). Header + polling badge counts subtract chat so the totals stay consistent — notifications themselves are untouched.
- **Places sheet** hides the Chat row (`show_chat: false`) while representing.

### The CI failure (the leak you flagged)
Fixed at the source: the header `user_menu` was re-rendering the **represented** identity's private-workspace link (that's the `private_workspace_test` failure) plus a **Sign Out** button that ends the underlying human login. Both are now hidden while representing; End Session in the banner stays. That test should pass now.

### One call left for you (D2 — menu contents)
I made the **minimal** menu change: hid only Workspace + Sign Out during representation, kept Profile / Help / Lists / Settings / AI Agents / Subdomains / Admin. Those remaining items point at the *represented* identity's pages (e.g. Settings → their settings). For agent representation that's benign; for human↔human representation you might want to trim more. Say the word and I'll pare it to a minimal Profile + Help set, or leave as-is.

Tests: added coverage for all four behaviors above; ran representation / private_workspace / chats / notifications / places-sheet suites + `srb tc` locally (docker) — all green.

danallison:
we should only hide chat for non-collective representation. for collectives, representation is the only way to access chat at all. same for user (profile) settings. representation is the only way for a collective to edit its public-facing profile. 

the problem is that it seems for both chat and user settings, navigating to `/chat` or `/settings` shows the representative's chat or settings and not the collective's, _but it does not actually end the representation session_. it _appears_ that the representation session has ended, but if you navigate to the collectives representation sessions list, it shows the session as still active. this is a separate bug that we should file an issue for, but for the purposes of this current pr, let's just hide chat and settings for all representation sessions.

to answer your question more directly, in any representation session, on the profile icon nav menu, we should **hide**:

* workspace link
* ai agents link (we might add this back later, but we need to think through the implications first)
* subdomains (cross-tenant representation sessions don't make sense)
* settings link
* admin link
* sign out button

so the only menu items that are visible in a representation session should be:

* profile
* lists
* help

melody-harmonic:
Reworked to your spec (`28e3ea98`) — and pulled the scope back in, since you're right that it was sprawling.

**Profile menu (desktop + mobile You tab), during *any* representation session** — trimmed to exactly:
- ✅ Profile · Lists · Help

and hidden:
- ❌ Workspace · AI Agents · Subdomains · Settings · Admin · Sign Out

**Chat + Settings** are now hidden at the **nav level only** for all representation sessions (chat row off the places sheet + its count zeroed so the places dot doesn't lie; Settings off the profile menu above).

**Reverted from my previous push (`3c369f2`)** — this is the "getting complicated" part you flagged:
- The `ChatsController` hard-block. Two reasons it was wrong: (1) it contradicts the end state you described — a collective reaches chat *only* via representation, so blocking chat for all representation would permanently wall that off; (2) my "DM leak" premise was mistaken. As you noted, `/chat` shows the **representative's own** chat and just leaves the session silently active — it doesn't expose the represented party's DMs. So there's nothing to *block* for safety here, only a confusing state to keep people out of, which nav-hiding does.
- The notifications-page chat filtering + header-count subtraction. You said hide chat specifically, not notifications — so notifications are back to untouched.

**Follow-up issue:** #419 — the real bug that navigating to `/chat` or `/settings` during representation shows your own surface and leaves the session active-but-apparently-ended. That's where the collective-vs-user nuance (chat/settings *routing through* a collective session) belongs.

**Tests:** swapped the chat-block/notification tests for a profile-menu trim test. `representation_test` + `places_sheet_component_test` + `private_workspace_test` + `api_collective_representation_test` all green locally (65 runs, 0 failures).

PR #420, opened 2026-07-06
Centralize representation enforcement as a before_action (#454, fixes #419)

### LINKED ISSUE #419: Navigating to /chat or /settings during representation silently leaves the session active while showing the representative's own surface
Split out from PR #417 (per @danallison's review on the note thread).

## Behavior
While a representation session is active, navigating directly to `/chat` or `/settings`:

- shows the **representative's own** chat / settings — **not** the represented party's, and not scoped through the representation; and
- **does not actually end the representation session.** It *appears* to have ended (you're looking at your own surfaces), but the collective's representation-sessions list still shows the session as active.

So the session is silently still running while the UI implies it isn't. That's confusing and a footgun — a representative can wander out of the represented context without realizing the session is still open.

## Scope of PR #417 (what's already handled)
#417 hides the `/chat` and `/settings` nav affordances during representation (chat row dropped from the places sheet; Settings + related items dropped from the profile menu), so users are far less likely to stumble into this state. It intentionally does **not** change controller behavior.

## The fix this issue tracks
Decide and implement the correct controller-level behavior when these routes are hit during representation. Per Dan's intended end state:

- **Chat:** for a **collective** representation session, representation is the *only* way to access the collective's chat, so `/chat` should route *through* the session (act as the collective) rather than dropping to the representative's own chat. For **non-collective (user)** representation, chat should stay hidden/blocked.
- **Settings:** representation is the only way for a **collective** to edit its public-facing profile, so `/settings` under a collective session should target the collective's settings. For user representation, settings should stay hidden/blocked.
- In all cases, hitting one of these routes must **not** silently leave a stale-but-active session behind: either keep the session applied to the route, or end it explicitly and reflect that in the sessions list — never the current in-between.

## Repro
1. Start a representation session (e.g. as a collective).
2. Navigate to `/chat` (or `/settings`).
3. Observe you see your own chat/settings, session appears ended.
4. Open the collective's representation-sessions list — the session is still active.

melody-harmonic:
Implements #454 (centralized representation policy layer) and fixes #419. Builds on the failing-test repro this branch started as — those tests are now the green acceptance gate.

Per Dan's call on #454: a collective may route through to its own /chat and /settings (act as itself, edit its public profile); user representation keeps the representative's personal surfaces blocked.

## Root cause (#419)
Representation path/route enforcement lived inside the **memoized** `current_representation_session` helper. Once a resolver (`resolve_browser_representation` / `resolve_api_representation`) had assigned `@current_representation_session`, the helper returned early and the guard below it never ran in the browser flow. So navigating directly to `/chat` or `/settings` while representing quietly rendered the representative's own personal surface while the session stayed **silently active** — the confusing "in-between" #419 flags.

## Change
New `RepresentationPolicy` concern runs `enforce_representation_scope!` as a **real `before_action`**, so no route can dodge it:

- **User representation** — blocks the representative's personal surfaces (`/chat`, `/settings`, matched as whole path segments so `/chat/:handle` and `/settings/two-factor` are covered) → bounce to `/representing`. The public space (`/`, `/about`, #383), the `/representing` dashboard, and collective-scoped pages render normally; the stop-representation flow (`DELETE /u/:handle/represent`) is unaffected.
- **Collective representation** — routes through: the collective may reach its own surfaces, including `/chat` and `/settings`.

The dead lazy guard is removed from `current_representation_session`, which is now a pure memoized reader.

`collectives#settings` / `update_settings` now authorize a collective's **own identity** (during an active collective representation session) to manage its own settings — its `identity_user` has no `CollectiveMember`, which is what made `is_admin?`-on-nil crash. Representing a collective is already a role-gated, reverified, audited action, so acting as the collective to edit its own profile is authorized. Scoped strictly to *this* collective's own identity.

## Scope / API note
Enforcement here is browser-only — API/MCP representation is validated in `resolve_api_representation` (session lookup + `X-Representing-*` headers) and signals failures with JSON errors, not redirects; #419 is a browser-nav bug. Grant collective-scope and private-workspace blocks stay in `validate_authenticated_access` (they must run before any spurious membership is created); this PR does not move them. Folding those into the policy object and unifying the two transport resolvers are natural follow-ups if you want the full consolidation — happy to do them once this shape is confirmed.

## Follow-up (your #454 reply)
"User settings sections should make sense per user type — human-only sections (2FA, devices, push, billing) only for humans." That's a distinct, representation-independent settings-UI concern; filing it as its own issue and linking here.

## Tests
`test/integration/representation_chat_settings_test.rb` (rewritten from failing-repro to acceptance): user-rep `/chat` & `/settings` → `/representing`; collective-rep `/settings` → 200 (own settings). Green locally, along with the browser + MCP representation suites (`representation_test`, `representation_session_test`, `api_representation_test`, `api_collective_representation_test`, `trustee_grant_flow_test`, `private_workspace_test`) and the `collectives` / `chats` / `users` controller tests.


danallison:
let's unify the resolver so that all representation goes through the same gate regardless of interface

melody-harmonic:
Done — unified the resolver in 4e49c4b9.

Both transports now route through one gate in `RepresentationPolicy`:
- **`representation_rejection_reason`** — the pure six-step validity sequence (exists → not ended → not expired → grant active → representative match → credential), returning a reason symbol or nil. `resolve_browser_representation` and `resolve_api_representation` both call it, so the checks can't drift between interfaces (the same class of gap #419 exposed on the enforcement side).
- **`apply_representation_session!`** — the shared "what a valid session does" (memoize, set context, swap `current_user`).
- Reason → surface maps (`API_REJECTION_MESSAGES` / `BROWSER_REPRESENTATION_FLASH`) keep every existing JSON error string and browser flash **verbatim**, so API clients and their tests are unchanged.

The transports now differ only at their I/O edges: session-id + credential source (cookie vs header) and rejection surface (clear+flash vs JSON 403). The X-Representing-* credential match keeps its own distinct, tested messages, so it stays a per-transport step but routes through the same last-check position.

No behavior change — all representation suites pass (api_representation, api_collective_representation, representation, representation_session, representation_chat_settings; 83 runs, 0 failures) and `srb tc` is clean.

app/controllers/concerns/representation_policy.rb
```diff
@@ -0,0 +1,139 @@
...
+#     may reach its own surfaces, including /chat and /settings, so it can act
+#     and edit its public profile as the collective.
+module RepresentationPolicy
+  extend ActiveSupport::Concern
+
+  # Top-level personal surfaces that a *user* representation must not render.
+  # Matched as whole path segments (the prefix itself or a "<prefix>/…"
+  # sub-path), so /chat, /chat/:handle, /settings and /settings/two-factor are
+  # all covered while unrelated paths like /chatter are not.
+  REPRESENTATIVE_PERSONAL_SURFACES = ["/chat", "/settings"].freeze
```
danallison, line 61: `/settings` redirects to `/u/<handle>/settings`, right? is that covered here?

now i'm wondering if maybe user settings _should_ live at `/settings` instead of `/u/<handle>/settings`, and if anything it should redirect in the other direction. a user's settings should only be accessible by that user, so it doesn't make much sense to specify the user handle in the path. no one should be visiting anyone else's user settings, so we don't need to differentiate between them in the path. if it makes it easier to enforce the representation rules implemented here, then i say we should swap the redirect direction so that `/u/<handle>/settings` redirects to `/settings` instead of the other way around.

melody-harmonic: Confirmed — real hole, not just a naming nit.

**Coverage question:** `/settings` is only a doormat — `users#redirect_to_settings` bounces it to `current_user.path + "/settings"` = `/u/<handle>/settings` (routes.rb:239, users_controller.rb:22). The `before_action` *does* catch bare `/settings` and redirect to `/representing` before the bounce fires, so that entry point is covered. But the **canonical** surface `/u/<handle>/settings` is not in `REPRESENTATIVE_PERSONAL_SURFACES`, so navigating there directly while representing still renders the representative's own settings. Same #419 dead-guard class, one layer down. So: partly covered, still leaky.

**On the swap:** I'm for it, and it's more than enforcement convenience. A handle-free canonical `/settings` collapses the blocklist to two literal top-level prefixes (`/chat`, `/settings`) with zero per-request handle interpolation — exactly the "one place, can't drift" property #454 is chasing. Keeping `/u/:handle/settings` canonical would instead force the guard to interpolate the *representative's* handle into the blocked path on every request, which is the scattered logic we're trying to delete.

**One wrinkle before I swap, because it dents the "settings are self-only" premise:** `/u/:handle/settings` isn't strictly self. `can_edit?` (user.rb:358) is `user == self || (user.ai_agent? && user.parent_id == id)` — so a human legitimately reaches **their own AI agent's** settings at `/u/<agent-handle>/settings`. That's the one place the `:handle` segment is load-bearing. Agents also already have `/ai-agents/:handle/settings` (routes.rb:60), so there's redundancy there.

So the swap needs one call from you on where agent settings land. My recommendation:
- `/settings` → the signed-in **human's own** settings (canonical, handle-free).
- `/u/:human-handle/settings` → 301 to `/settings` (keeps old links/bookmarks/emails alive).
- Agent settings stay at the existing `/ai-agents/:handle/settings`; `/u/:agent-handle/settings` redirects there.
- `/chat` + `/settings` become the whole blocklist; `/ai-agents/:handle/settings` is inherently handle-scoped (you manage N agents) and stays out of it.

If that matches your intent, I'll do the route move + redirect flips + path-helper/view sweep (~160 `/settings` refs) on this branch and re-push. If you'd rather I plug the immediate hole first (also block the representative's own `/u/<their-handle>/settings`) and do the swap as a separate PR, say so — but I think the swap is the correct single fix and I'd rather not ship the interpolation-based guard we're trying to kill.

app/controllers/concerns/representation_policy.rb
```diff
@@ -0,0 +1,139 @@
...
+    return unless session.user_representation?
+
+    return unless representative_personal_surface?(request.path)
+
+    redirect_to "/representing"
+  end
+
+  def representative_personal_surface?(path)
+    REPRESENTATIVE_PERSONAL_SURFACES.any? do |prefix|
+      path == prefix || path.start_with?("#{prefix}/")
```
danallison, line 136: does this cover query strings like `/settings?q=1`, or is the query string assumed to be stripped from the path already? we should have a test case to confirm.

app/controllers/concerns/representation_policy.rb
danallison: that sounds good. go ahead and make the swap now.

app/controllers/concerns/representation_policy.rb
melody-harmonic: Confirmed and covered — the query string is already stripped. The guard matches `request.path` (Rack's path without the query string; `request.fullpath` is the one that keeps it), so `/settings?x=1` matches `/settings` exactly.

Added a test in `representation_chat_settings_test.rb` that, while representing, hits `/settings?foo=1&bar=2` and `/chat?tab=dms` and asserts both still bounce to `/representing` — so a future refactor that switched to `fullpath` can't silently reopen the hole. Pushed in bf632784.

app/controllers/concerns/representation_policy.rb
melody-harmonic: Done — pushed bf632784. `/settings` is now the canonical, handle-free surface; `/u/:handle/settings(/*rest)` 308-redirects to it (humans) or to `/ai-agents/:handle/settings` (agents). The blocklist stays the two literal prefixes `/chat` + `/settings`, with zero per-request handle interpolation.

The settings controllers default `params[:handle]` to `current_user` via a small `SettingsSubjectDefaulting` concern, so their action bodies are unchanged.

**Two load-bearing exceptions I kept handle-scoped** (flagged so you can veto):
- **API tokens** (`/u/:handle/settings/tokens`) — an agent owns its own tokens and manages them from its `/ai-agents/:handle/settings` page, so the `:handle` identifies the owner (you *or* one of your agents). That surface is `can_edit?`-gated, so it isn't a representation hole.
- **Email-change confirm** (`/u/:handle/settings/email/confirm/:token`) — a token-authenticated deep link that must resolve the account with **no session** (clicked from an email in any browser), so the handle, not `current_user`, names whose pending email is being confirmed.

Everything else moved: profile, email change, workspace-trio, devices, push-subscriptions, notifications, data-export, and **trustee-authorizations**. Trustee moved too because `authorize_grant_management` already required `target == current_user`, so the handle was never load-bearing in the browser; the notification/warning-path builders and the agent capability-index keys moved with it.

`srb tc` clean; the affected suites are green locally (representation, users, devices, web-push, data-export, trustee, api-tokens, api-representation, the actions_helper capability index, and the mailer route-recognition tests). The cross-user authorization tests that are now impossible-by-construction were reframed to pin the new "handle-free route only ever touches the signed-in user" invariant rather than deleted.

PR #426, opened 2026-07-06
Make @ optional in the collective: search filter (#356)

### LINKED ISSUE #356: make @ optional in handle seach filters
there's inconsistency with the handle search filters. collectives use `collective:handle` but users use `creator:@handle`. we should standardize on `:handle` without the `@`, but still allow the `@` so that the search results are the same either way.

melody-harmonic:
Closes #356.

**Inconsistency:** user handle filters (`creator:`, `read-by:`, `voter:`, `participant:`, `mentions:`, `replying-to:`) already accept an optional leading `@` and strip it (`build_user_filter_param`), so `creator:@alice` and `creator:alice` return identical results. But `collective:` didn't — its value pattern rejected `@`, so `collective:@my-team` fell through to a fuzzy text term and returned different (wrong) results than `collective:my-team`.

**Fix:** allow an optional leading `@` in `COLLECTIVE_HANDLE_PATTERN` and strip it in `build_collective_param`, bringing `collective:` to parity with the user filters. Now every handle filter is `@`-optional and "the same either way."

Docs: the search help now presents the bare-handle form as canonical (`creator:alice`) and notes the `@` is optional across all handle filters.

Tests: added `collective:@handle` parse cases (including parity with the no-`@` form). `test/services/search_query_parser_test.rb` green locally (118 runs, 0 failures); `srb tc` clean.

### DIFF
--- a/app/services/search_query_parser.rb
+++ b/app/services/search_query_parser.rb
@@ -15,7 +15,9 @@ class SearchQueryParser
   HANDLE_PATTERN = /^@?[a-zA-Z0-9_-]+$/
 
   # Collective handle pattern (alphanumeric with dashes)
-  COLLECTIVE_HANDLE_PATTERN = /^[a-zA-Z0-9-]+$/i
+  # Leading @ is optional so collective:@my-team and collective:my-team behave
+  # identically, matching the user handle filters (issue #356).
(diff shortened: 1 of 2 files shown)


danallison:
let's document the filters as if `@` is assumed. we don't need to explicitly call out that it is optional. it's just an implicit flexibility to accommodate either format.

PR #437, opened 2026-07-06
Let free accounts buy LLM credits without a subscription (#433)

### LINKED ISSUE #433: bug: free accounts (app admins) cannot add llm credits from the billing page
free accounts do not have to pay the $3/month fee for ai agent identities or collective automations, but llm credits still cost money, so free accounts still need to have the option to purchase llm credits to power internal agents.

melody-harmonic:
Closes #433.

Free accounts (app admins whose AI-agent/collective resources are billing-exempt) don't pay the $3/mo fee, so they never open a Stripe subscription. But LLM credits are a **separate one-time purchase** that powers internal agents — and both the billing page and the `topup` action required an *active subscription*, so free accounts had no way to add credits.

## Fix
- **`BillingController#topup`** — gate on `stripe_billing` being enabled instead of an active subscription, and `find_or_create_customer` so a free account with no prior subscription can attach a one-time credit payment. (Errors when billing is disabled.)
- **`BillingController#load_credit_balance`** — surface the balance whenever billing is enabled; a free account with no Stripe customer yet shows `$0` until its first top-up creates one.
- **View** — extracted the "AI Agent Credits" section into `billing/_credits_section` and rendered it in the free/exempt branch (`billable_quantity == 0`), not only the active-subscription branch.

Credit grants already call `ensure_pricing_plan_subscription!`, so a free account's purchased credits meter and drain correctly once bought — no separate subscription needed.

## Tests
`billing_controller_test`:
- the billing page offers the top-up form to a free account
- a free account can start a credit top-up (reaches Stripe Checkout) with no active subscription
- topup is rejected when `stripe_billing` is disabled

### DIFF
--- a/app/controllers/billing_controller.rb
+++ b/app/controllers/billing_controller.rb
@@ -77,8 +77,12 @@ class BillingController < ApplicationController
 
   # POST /billing/topup
   def topup
-    unless current_user.stripe_customer&.active?
-      flash[:error] = "You need an active subscription before adding credits."
+    # Credits are a one-time payment, independent of the $3/mo subscription.
+    # Free accounts (all resources billing-exempt, no subscription) still need
+    # to buy LLM credits to power internal agents — so gate on billing being
+    # enabled, not on an active subscription.
+    unless current_tenant&.feature_enabled?("stripe_billing")
+      flash[:error] = "Billing is not enabled."
       return redirect_to billing_show_path
     end
 
@@ -95,7 +99,9 @@ class BillingController < ApplicationController
       return redirect_to billing_show_path
     end
 
-    stripe_customer = current_user.stripe_customer
+    # find_or_create so a free account with no prior subscription can still
+    # attach a one-time credit payment to a Stripe customer.
+    stripe_customer = StripeService.find_or_create_customer(current_user)
     billing_url = billing_show_url
     success_url = "#{billing_url}?checkout_session_id={CHECKOUT_SESSION_ID}"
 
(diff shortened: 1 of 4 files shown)


app/controllers/billing_controller.rb
```diff
@@ -77,8 +77,12 @@ def portal
...
   def topup
-    unless current_user.stripe_customer&.active?
-      flash[:error] = "You need an active subscription before adding credits."
+    # Credits are a one-time payment, independent of the $3/mo subscription.
+    # Free accounts (all resources billing-exempt, no subscription) still need
+    # to buy LLM credits to power internal agents — so gate on billing being
+    # enabled, not on an active subscription.
+    unless current_tenant&.feature_enabled?("stripe_billing")
+      flash[:error] = "Billing is not enabled."
       return redirect_to billing_show_path
```
danallison: this doesn't make sense. if billing is not enabled, why redirect to the billing page? please think carefully about the full end-to-end flow in each of the two distinct scenarios (free account vs not free account), and make sure every step makes sense.

melody-harmonic: You're right — walking both flows end to end, that redirect never made sense:

- **Free account** (`billable_quantity == 0`): lives on a tenant with `stripe_billing` **enabled**. Form renders → topup → guard passes → `find_or_create_customer` → checkout. ✓
- **Not free** (`billable_quantity > 0`): also `stripe_billing` **enabled**. Active sub → form → topup → checkout; no sub yet → "Subscription required", no top-up form. ✓

So in **both** real scenarios billing is enabled — "free account" isn't "billing disabled," it's a billing-exempt user *inside* a billing-enabled tenant. The `unless feature_enabled?` branch is therefore a defensive path that the UI never reaches (the form only renders when the flag is on), and sending it to `/billing` was doubly wrong: that page is itself empty/broken when billing is off.

I kept the guard — Stripe keys are app-global, so without it a hand-crafted POST could open a real checkout on a tenant where billing is deliberately off — but changed the destination to `root_path` (pushed 1 commit; test updated). If you'd rather I harden the *whole* BillingController for the billing-disabled case (a `before_action` covering show/setup/portal too, which today also render incoherently when the flag is off), say the word and I'll do it as a follow-up — it's out of #433's scope but the same root cause.

PR #443, opened 2026-07-07
Show per-cycle check-in hearts in the places sheet (#418)

### LINKED ISSUE #418: collectives in the place sheet should show the same heartbeat indicators as `/collectives`
`/collectives` shows a heart next to each collective that is either empty or filled depending on whether the current user has sent a heartbeat to the collective in the current cycle. the places sheet should do the same thing (but keep the alphabetical sorting on the places sheet).

melody-harmonic:
Closes #418.

## What
`/collectives` shows a filled/empty heart next to each collective for whether the current user has submitted a heartbeat this cycle. The places sheet (the switcher opened from the header's places toggle and the mobile Places tab) listed collectives with only avatar + name + unread badge. This adds the same heart to the sheet, **keeping the sheet's alphabetical sort** (as the issue requested — no re-sort by check-in state).

## How
- **Single source for the heartbeat lookup.** Extracted a `Collective.with_heartbeat_for(user)` scope and reused it in both `CollectivesController#index` and `ApplicationController#places_collectives`, replacing the inline `LEFT JOIN heartbeats … / SELECT … IS NOT NULL AS has_heartbeat` that previously lived only in the controller.
- **EXISTS, not JOIN.** The scope uses an `EXISTS` subquery rather than a LEFT JOIN so it can never multiply collective rows. The places sheet renders on *every* page, so a row-duplication bug would be pervasive; EXISTS sidesteps it entirely. `has_heartbeat` remains an orderable boolean, so `/collectives`' existing `.order(:has_heartbeat, :name)` is unchanged.
- **Safe default.** `Collective#has_heartbeat` now returns `false` when the attribute wasn't selected, so any view can call it on a plainly-loaded record without raising.
- **Placement.** The heart sits between the name and the unread badge, using the existing `pulse-heartbeat-icon` / `-sent` / `-pending` classes. The public-space globe row is not a check-in collective, so it stays heartless — matching `/collectives`, which excludes the main collective.

## Tests
- New component test asserts the filled heart for a checked-in collective, the empty heart for a pending one, and no heart on the globe row.
- `places_sheet_component_test.rb` + `collectives_controller_test.rb` green locally (the controller test exercises the new scope against Postgres). `srb tc` clean.

### DIFF
--- a/app/assets/stylesheets/pulse/_places_sheet.css
+++ b/app/assets/stylesheets/pulse/_places_sheet.css
@@ -179,23 +179,12 @@
   flex-shrink: 0;
 }
 
-/* Per-cycle check-in heart, mirroring the /collectives list. Sits between the
-   name and the unread badge; never shrinks. */
+/* Per-cycle check-in heart, mirroring the /collectives list. Always present, so
+   it's the trailing column: the unread badge sits to its left and collapses
+   (display:none) when zero, letting that column's width vary without disturbing
+   the heart's stable right-aligned column. Never shrinks. */
 .pulse-places-heartbeat {
   flex-shrink: 0;
-}
-
-/* Fixed-width trailing slot for the unread badge on collective rows. The badge
-   collapses (display:none) when the count is zero, so without a reserved slot
-   the heart would slide right into the badge's column on checked-in rows with
-   nothing unread. The slot keeps the heart in a stable column and right-aligns
-   the badge against the row edge whether or not it's shown. */
-.pulse-places-badge-slot {
-  display: flex;
-  align-items: center;
-  justify-content: flex-end;
-  flex-shrink: 0;
-  width: 28px;
 }
 
 .pulse-places-row-add {
--- a/app/assets/stylesheets/pulse/_places_sheet.css
+++ b/app/assets/stylesheets/pulse/_places_sheet.css
@@ -185,6 +185,19 @@
   flex-shrink: 0;
 }
 
+/* Fixed-width trailing slot for the unread badge on collective rows. The badge
+   collapses (display:none) when the count is zero, so without a reserved slot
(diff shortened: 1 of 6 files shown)


danallison:
when there's an unread count badge, that displaces the heart such that it breaks the heart column. iow, it looks like this

|                   |              |              |
|------------|--------|---------|
| collective a |             | [heart] |
| collective b | [heart] | [unread] |

when it should look like this

|                   |              |              |
|------------|--------|---------|
| collective a | [heart] |              |
| collective b | [heart] | [unread] |

melody-harmonic:
Good catch — fixed in a3ff65a2. The unread badge collapses to `display:none` when the count is zero, so on a checked-in row with nothing unread the heart slid right into the badge's column. Wrapped the badge in a fixed-width slot (`.pulse-places-badge-slot`), so the heart now stays in a stable column and the badge right-aligns against the row edge whether or not it's shown — exactly your second diagram. Added a component-test assertion that the badge lives inside the slot; suite green locally.

danallison:
problem now is that there's a weird gap/margin whenever there's no unread badges. instead of having two fixed width columns, let's swap the column order so that hearts are always on the right and unread badges appear to the left of the hearts. hearts are always present, so that column will always have the same consistent width, but unread badges are only sometimes present, so keeping them to the left of the hearts allows the width to vary without breaking the right-aligned hearts column.

melody-harmonic:
Done in 1e3441c3. Swapped the column order — the unread badge now sits to the left of the heart, and I dropped the fixed 28px slot. Since the heart is always rendered it forms a stable right-aligned column, so a hidden badge (zero count) just lets its column collapse without any leftover gap. Component test updated to assert the heart is the badge's immediate next sibling; `bin/rails test test/components/places_sheet_component_test.rb` green (9 runs, 0 failures).

PR #452, opened 2026-07-07
Fund internal agents with prepaid credits, not workspace subscription (#450)

### LINKED ISSUE #450: bug: free accounts can't run internal agents
https://github.com/ibis-coordination/Harmonic/pull/437 fixed the issue where free accounts could not add llm credits. but now there's a second issue which is that those llm credits can't be used because the users is categorized as not having stripe set up yet. attempting to chat with an agent results in

> Billing is not set up. Please set up billing at /billing before running AI agents.

even though there are purchased llm credits available.

melody-harmonic:
Fixes #450.

## Problem

#437 let free accounts (no paid workspace subscription) buy LLM credits. But the agent-dispatch billing gate still required an **active workspace subscription** (`StripeCustomer#active?`), so a free account with purchased credits hit:

> Billing is not set up. Please set up billing at /billing before running AI agents.

…and could never spend the credits it had bought.

## Root cause

AI usage is funded by the **metered pricing-plan subscription** (`pricing_plan_subscription_id`), which any credit top-up sets up (`StripeService.ensure_pricing_plan_subscription!`). That is *decoupled* from the paid workspace subscription (`active?`). The gates conflated the two.

## Fix

Gate agent usage on the pricing-plan subscription (prepaid AI credits), not `active?`:

- **`AgentRunnerDispatchService`** — block only when there's no pricing-plan subscription, and route to the Stripe gateway on that same signal so a free account's credits actually drain (previously routed to LiteLLM). The in-gateway pricing-plan recheck is now guaranteed by the entry gate, so it's removed.
- **`Internal::AgentRunnerController#preflight`** — same gate, so the runner doesn't re-block a task that dispatch already allowed.
- **`AutomationExecutor`** — same gate on both the rule-run and `trigger_agent` paths, keeping scheduled/triggered automations consistent with chat.

An active paid account that has never topped up still cannot run agents (no pricing-plan subscription) — behavior unchanged, just a clearer message ("AI usage billing is not set up. Add credits at /billing").

## Tests

- New regression: a free account (`active: false`) with a pricing-plan subscription + credits **dispatches** through `stripe_gateway`.
- Updated the renamed "not active" gate test to assert the AI-usage message.
- Full `agent_runner_dispatch_service`, `automation_executor`, and `internal/agent_runner_controller` suites green (108 runs, 0 failures); `srb tc` clean.

### DIFF
--- a/app/controllers/internal/agent_runner_controller.rb
+++ b/app/controllers/internal/agent_runner_controller.rb
@@ -289,8 +289,11 @@ module Internal
       # parallel skip in AgentRunnerDispatchService.
       if tenant.feature_enabled?("stripe_billing") && !ai_agent.system?
         billing_customer = ai_agent.billing_customer
-        unless billing_customer&.active?
-          render json: { status: "fail", reason: "Billing is not set up" }
+        # Prepaid AI credits (the metered pricing-plan subscription), not the
+        # paid workspace subscription (active?), fund agent usage — a free
+        # account with LLM credits can run agents. Mirrors the dispatch gate.
+        if billing_customer.nil? || billing_customer.pricing_plan_subscription_id.blank?
+          render json: { status: "fail", reason: "AI usage billing is not set up" }
           return
         end
 
--- a/app/controllers/internal/agent_runner_controller.rb
+++ b/app/controllers/internal/agent_runner_controller.rb
@@ -289,9 +289,20 @@ module Internal
       # parallel skip in AgentRunnerDispatchService.
       if tenant.feature_enabled?("stripe_billing") && !ai_agent.system?
         billing_customer = ai_agent.billing_customer
-        # Prepaid AI credits (the metered pricing-plan subscription), not the
-        # paid workspace subscription (active?), fund agent usage — a free
-        # account with LLM credits can run agents. Mirrors the dispatch gate.
+
+        # (a) The agent's identity must be paid for — an active billing_customer
+        # (its principal's per-identity subscription) — unless the principal is a
+        # free account (nothing billable, e.g. an app admin), which owes no such
+        # fee. billable_quantity of zero is exactly that case. Mirrors the
+        # dispatch gate.
+        unless billing_customer&.active? || ai_agent.parent&.billable_quantity&.zero?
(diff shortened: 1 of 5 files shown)


danallison:
this seems confused. what does a "workspace account" mean? i don't know what that is. we should still check to make sure that each agent's identity is being paid for before we allow task runs. that shouldn't change. the only exception is if the principal human user has an exemption (free account). the exemption is a special case and not the norm.

melody-harmonic:
Reworked in 4acc43aa per your correction — dropped the invented "workspace subscription/account" framing entirely.

The confusion was that I'd collapsed two independent requirements into one. They're now both explicit:

- **(a) Identity paid for** — the agent's `billing_customer` (which *is* its principal's Stripe customer, via `assign_billing_customer!`) holds an active per-identity subscription. **This is the norm and it stays**, exactly as you said. The one exception is a free-account principal, applied as a special case.
- **(b) LLM usage funded** — a prepaid-credit (pricing-plan) subscription with a positive balance. Required for **every** billed agent, free or paying. This is the actual #450 fix: a free account that bought credits (`active? == false`, pricing-plan present) now clears (a) via the exemption and (b) via its credits, and gateway routing no longer keys on `active?` so those credits actually drain through the Stripe AI Gateway.

Applied to the dispatch service, both automation gates, and preflight.

**One correction to what I told you on the note thread:** I said I'd gate the exemption on `parent.billing_exempt?`. Digging into #437, "free account" is more precisely *"the principal owes no per-identity fee"* — i.e. `principal.billable_quantity == 0` (an app admin, or a principal whose resources are all billing-exempt). That's strictly broader than the `billing_exempt` flag alone, which would miss app-admin free accounts — the exact case #437 describes. So the gate is `billing_customer&.active? || principal&.billable_quantity&.zero?`. If you'd rather it be narrower (e.g. only an explicit flag), easy one-line change — flag it.

Tests: reworked the regression to the real scenario (app-admin principal + prepaid credits → dispatches via stripe_gateway), added a gate-(b) case (free account, no credits → fails), fixed the identity-gate message. 109 runs green across dispatch/automation/preflight; `srb tc` clean.

PR #453, opened 2026-07-07
Special @-tags: @everyone / @admins + reserved-handle registry (#449)

### LINKED ISSUE #449: special @ tags for in-collective roles and aliases
tags like `@admin` or `@everyone`. note that this also has relevance for ai agents that instantiate in multiple collectives as distinct user records while all sharing the same name like `@trio` or `@claude`.

melody-harmonic:
Closes #449. Implements **Option B** from the design decision (harmonic-devs [ddeb6975](https://www.harmonic.social/collectives/harmonic-devs/d/ddeb6975), both voters starred it).

## What this does

Adds two collective-local group mention tags and folds handle-reservation into one registry.

### `ReservedHandles` registry (`app/models/reserved_handles.rb`)
Single source of truth for handles that aren't ordinary, user-claimable handles. Folds in the three places this fact used to live — `Collective::RESERVED_HANDLES`, `TenantUser::RESERVED_HANDLES`, `MentionParser::TRIO_HANDLE` — which had already drifted (a collective could be named `trio` even though the handle was reserved for users). It distinguishes:
- **Group tags** (`@everyone`, `@admins`) — expand to a *set* of users; never a real record, so no user or collective may claim them.
- **Agent handles** (`@trio`) — real records sharing one name across collectives, each resolving collective-locally. Role-gated. This is also the mechanism for the cross-collective agent-identity angle Dan flagged (`@claude`/`@trio` → local instance).

### Group-mention resolution (`MentionParser`)
- `@admins` → the collective's admins. Any member may use it.
- `@everyone` → all members, **but only when the author is an admin**. Without a known admin author (background/system parse paths) it expands to nobody, so it can never fan out by accident. The author is now threaded through the notification dispatcher and the reminder path.
- Results are deduped, so a user named directly *and* via a tag is notified once.
- Both resolve **collective-locally** — an `@everyone` in one collective never reaches another.
- Render path links `@everyone`/`@admins` to the collective page (display only; the admin gate lives on the notification path).

### Reservation
`everyone`/`admins` are now reserved across the unified user+collective namespace, so neither a user nor a collective identity can shadow the tag. Agent handles stay claimable as *collective* handles (backwards compatible; only the group tags are newly reserved).

## Deferred (follow-ups, not in this PR)
- **`@here`** (active/present members) — explicitly out of scope for v1 per Option B.
- **Interactive confirm dialog** for large `@everyone` fan-outs — the admin-only gate is the guardrail in this pass; a compose-time "notify N members?" confirm is a UI-layer follow-up.
- **Autocomplete** — surfacing `@everyone`/`@admins` in the mention autocomplete dropdown (the tags work when typed; they're just not suggested yet).

## Testing
All local, via `docker-compose.dev.yml`:
- `reserved_handles_test.rb` (new), `mention_parser_test.rb` (+group-tag section), `tenant_user_test.rb`, `collective_test.rb`, `notification_dispatcher_test.rb`, `automation_mention_filter_test.rb`, `markdown_renderer_test.rb`, `autocomplete_controller_test.rb` — all green.
- `srb tc` — no errors.

No migration (all resolution logic + a Ruby registry), so no `structure.sql` change.

### DIFF
--- a/app/views/help/notifications.md.erb
+++ b/app/views/help/notifications.md.erb
@@ -16,6 +16,15 @@ Notifications keep you aware of activity that involves you across all collective
 
 Each notification links back to the content or event that triggered it.
 
+## Group Mentions
+
+Alongside mentioning an individual by their `@handle`, two special tags notify a group within the collective they're written in:
+
+- **`@admins`** — notifies the collective's admins. Any member may use it.
+- **`@everyone`** — notifies every member of the collective. Only admins can trigger an `@everyone` fan-out; written by a non-admin it's shown but notifies no one.
+
+Both resolve **collective-locally**: an `@everyone` in one collective never reaches another. These handles are reserved, so no one can register a user or collective named `everyone` or `admins`.
+
 ## Unread, Read, and Dismissed
 
 Notifications move through three states:
--- a/app/views/help/notifications.md.erb
+++ b/app/views/help/notifications.md.erb
@@ -18,12 +18,19 @@ Each notification links back to the content or event that triggered it.
 
 ## Group Mentions
 
-Alongside mentioning an individual by their `@handle`, two special tags notify a group within the collective they're written in:
+Alongside mentioning an individual by their `@handle`, a few special tags notify a group within the collective they're written in.
 
-- **`@admins`** — notifies the collective's admins. Any member may use it.
-- **`@everyone`** — notifies every member of the collective. Only admins can trigger an `@everyone` fan-out; written by a non-admin it's shown but notifies no one.
+**Role tags** notify every member holding a [collective role](/help/collectives#roles):
 
-Both resolve **collective-locally**: an `@everyone` in one collective never reaches another. These handles are reserved, so no one can register a user or collective named `everyone` or `admins`.
+- **`@admins`** — notifies the collective's admins.
+- **`@representatives`** — notifies its representatives.
+- **`@summarizers`** — notifies its summarizers.
+
+Any member may use a role tag. The set of role tags follows the collective's role list, so every role has a matching `@<role>s` tag — if a role is added later, its tag works automatically.
+
+**`@everyone`** notifies every member of the collective. Only admins can trigger an `@everyone` fan-out; written by a non-admin it's shown but notifies no one.
+
+All of these resolve **collective-locally**: an `@everyone` in one collective never reaches another. The tags are reserved, so no one can register a user or collective that would shadow one.
 
 ## Unread, Read, and Dismissed
 
--- a/app/models/collective.rb
+++ b/app/models/collective.rb
@@ -883,14 +883,23 @@ class Collective < ApplicationRecord
     !!(open_to_all || all_members_can_invite)
   end
 
+  # Members holding `role` (e.g. "admin", "representative", "summarizer").
+  # Backs the role group tags (@admins/@representatives/… — see
+  # MentionParser.resolve_collective_local), which are derived from the role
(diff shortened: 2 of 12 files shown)


danallison (changes requested):
we should have handles for all collective roles including `representatives` and `summarizers`. eventually there will likely also be custom roles, so if it's simple enough to do, we should derive the list of reserved role handles from the collective's config (but if that introduces any new complexity at all then just skip it).

melody-harmonic:
Done — pushed. Role group tags are now **derived from the collective role list** (`CollectiveMember.valid_roles`) rather than hardcoded, so this covers `@admins`, `@representatives`, and `@summarizers`, and any role added later (including future per-collective custom roles) automatically gets its `@<role>s` tag reserved and resolved with no change to the registry.

- `ReservedHandles.role_tags` maps each role → its pluralized tag (`admin`→`@admins`, `representative`→`@representatives`, `summarizer`→`@summarizers`); `group_tags` = `@everyone` + those. Reservation (`forbidden_for_user?`/`forbidden_for_collective?`) and collective-local resolution both read from it.
- `MentionParser` expands every role tag to its role holders (any member may use a role tag, same as `@admins`); the render path already links them all via `group_tag?`.
- `Collective#users_with_role` backs the lookup; `admins`/`representatives` route through it.

The derivation was simple (roles already have a single source of truth in `valid_roles`), so it introduced no new complexity — the "skip if complex" escape hatch wasn't needed. `@everyone`'s admin-only gate is unchanged.

Local: `reserved_handles_test` + `mention_parser_test` (new role-tag cases) + `collective_test` + `tenant_user_test` green (213 runs, 0 failures); `srb tc` clean.

app/views/help/notifications.md.erb
```diff
@@ -16,6 +16,15 @@ Notifications keep you aware of activity that involves you across all collective
...
 Each notification links back to the content or event that triggered it.
 
+## Group Mentions
+
+Alongside mentioning an individual by their `@handle`, two special tags notify a group within the collective they're written in:
+
+- **`@admins`** — notifies the collective's admins. Any member may use it.
+- **`@everyone`** — notifies every member of the collective. Only admins can trigger an `@everyone` fan-out; written by a non-admin it's shown but notifies no one.
+
+Both resolve **collective-locally**: an `@everyone` in one collective never reaches another. These handles are reserved, so no one can register a user or collective named `everyone` or `admins`.
```
danallison: this page needs updating following the addition of the other roles (`representatives`, `summarizers`)

melody-harmonic: Done in 8fd89668. Rewrote the Group Mentions section to document the full role-tag family: `@admins`, `@representatives`, `@summarizers` (each notifies members holding that [collective role](/help/collectives#roles), any member may use a role tag), alongside the admin-gated `@everyone`. Also noted that the tag set follows the role list, so a role added later gets its `@<role>s` tag automatically — matching how `ReservedHandles.role_tags` derives them from `CollectiveMember.valid_roles`.

Separate gap I noticed while here: the `/help/collectives` **Roles** section lists Admin / Representative / Member but not **Summarizer**, even though the role exists in code and now has a `@summarizers` tag. Want me to add Summarizer there in this PR, or leave it for a docs pass?

danallison: let's add it now

melody-harmonic: Done in 408e5dc3 — added **Summarizer** to the `/help/collectives` Roles list ("Can write and maintain the summaries of the collective's content"), placed between Representative and Member.

Issue #345, opened 2026-07-02
collective identity user notifications routing/visibility


danallison:
since collectives can act as individual identities, there needs to be a clear way for representatives to respond to notifications _for the collective_. if a non-member @ mentions the collective externally, the collective itself should get a notification, but currently there's no visibility or routing for that. so representatives don't know about it unless they happen to start a representation session and see the notifications.

PR #475, opened 2026-07-09
Fix #468: allow tuning in to a collective-identity user

### LINKED ISSUE #468: bug: can't tune in to collective identity user
the POST route used is `/collectives/<handle>/actions/tune_in` instead of `/u/<handle>/actions/tune_in`, resulting in 404

melody-harmonic:
Fixes #468.

## Problem

Tuning in to a collective-identity user failed. Dan reported the visible symptom: the button POSTed to `/collectives/<handle>/actions/tune_in` and 404'd. Behind that 404 was a second wall — even on the correct route the action returned 422.

## Root causes & fixes

**1. Wrong route (the reported 404).** `app/views/shared/_tune_in_button.html.erb` built its URL from `target.path`. For a collective-identity user `User#path` points at the collective profile (`/collectives/<handle>`), which has no `tune_in` action. The button now targets the user resource (`/u/<handle>/actions/tune_in`) using the **TenantUser** handle — note `User#handle` returns the `"collectives/<handle>"` display form for identity users, not a routable handle, so we read `target.tenant_user&.handle` (the suffixed, resolvable handle that `/u/:handle` looks up).

**2. Membership validation (the hidden 422).** `execute_tune_in` adds the target to the actor's **main-collective-scoped** primary list, but `UserListMember#member_is_collective_member` rejected it: collective-identity users are structurally barred from main-collective *membership* (a collective can't be a member of the tenant-wide "everyone"). They're still tenant citizens you can tune in to, so the validation now admits an identity user onto a main-collective list when its collective lives in the list's tenant. Non-main lists stay strict — an identity user must be an actual member.

This partial is shared by the user-lists, mutuals, and notifications surfaces, so all three are fixed.

## Tests
- `user_lists_html_test` — the members-tab tune-in button for an identity member targets `/u/<handle>`, never the collective path or the `collectives/` display form.
- `users_tune_in_actions_test` — full tune-in flow to a collective identity succeeds (200) and the identity lands on the primary list.
- `user_list_member_test` — the main-collective exception does **not** leak: an identity non-member is still rejected from a non-main list.

Ran the three files + `notifications_controller_test` + `srb tc` locally: all green.

### DIFF
--- a/app/models/user_list_member.rb
+++ b/app/models/user_list_member.rb
@@ -63,7 +63,26 @@ class UserListMember < ApplicationRecord
   sig { void }
   def member_is_collective_member
     return if CollectiveMember.exists?(collective_id: collective_id, user_id: user_id)
+    return if addable_collective_identity?
 
     errors.add(:user_id, "must be a member of the collective")
+  end
+
+  # Collective-identity users are structurally barred from *membership* in the
+  # main collective — a collective can't be a member of the tenant-wide
+  # "everyone" — but they're still tenant citizens you can tune in to. The
+  # primary "tuned in" list is main-collective-scoped, so admit an identity
+  # user there as long as its collective lives in this list's tenant. Without
+  # this, tuning in to a collective identity fails validation even once the
+  # request reaches the right route (issue #468).
+  sig { returns(T::Boolean) }
+  def addable_collective_identity?
+    member = user
+    list = user_list
+    return false if member.nil? || list.nil?
+    return false unless member.collective_identity?
+    return false unless list.collective&.is_main_collective?
+
+    Collective.where(identity_user_id: member.id, tenant_id: list.tenant_id).exists?
   end
 end
--- a/app/views/shared/_tune_in_button.html.erb
+++ b/app/views/shared/_tune_in_button.html.erb
@@ -8,12 +8,21 @@
   # `tune_in_on_list_user_ids` for the batch-loading pattern.
 %>
 <% if viewer && viewer.id != target.id && !blocked %>
+  <%
+    # Tune-in acts on the user resource at /u/<handle>, NOT target.path: for a
+    # collective-identity user `#path` points at the collective profile
+    # (/collectives/<handle>), which has no tune_in action — so appending it
+    # 404s (issue #468). Use the TenantUser handle (which the /u/:handle routes
+    # resolve via find_by), NOT User#handle — the latter returns the display
+    # form "collectives/<handle>" for identity users, not a resolvable handle.
+    user_action_base = "/u/#{target.tenant_user&.handle}"
+  %>
   <%= render AjaxToggleButtonComponent.new(
         on:        on_list,
-        on_url:    "#{target.path}/actions/tune_out",
+        on_url:    "#{user_action_base}/actions/tune_out",
         on_html:   pulse_icon('tuning_in', size: :sm) + " Tuned in",
         on_class:  "pulse-action-btn-secondary pulse-tune-in-btn",
-        off_url:   "#{target.path}/actions/tune_in",
+        off_url:   "#{user_action_base}/actions/tune_in",
         off_html:  octicon('plus', height: 14) + " Tune in",
         off_class: "pulse-action-btn pulse-tune-in-btn",
         title:     "Tuning in adds this user's activity to your timeline view",
--- a/test/controllers/user_lists_html_test.rb
+++ b/test/controllers/user_lists_html_test.rb
@@ -194,6 +194,43 @@ class UserListsHtmlTest < ActionDispatch::IntegrationTest
     assert_no_match(/pulse-tune-in-btn/, own_row.to_s)
   end
 
+  test "show HTML: tune-in button for a collective-identity member targets /u/<handle>, not the collective path (issue #468)" do
+    # A standard collective auto-creates an identity User that shares the
+    # collective's handle and owns a TenantUser — so it is a real, tune-in-able
+    # user reachable at /u/<handle>. Its #path, however, points at the
+    # collective profile (/collectives/<handle>), which has no tune_in action.
+    # The button must build its URL from the user resource, not from #path,
+    # else it POSTs to a 404.
+    identity_collective = Collective.create!(
+      tenant: @tenant, name: "Ident Co", handle: "ident-#{SecureRandom.hex(4)}",
+      collective_type: "standard", created_by: @user, updated_by: @user
+    )
+    identity_user = identity_collective.identity_user
+    assert identity_user, "expected the standard collective to have an identity user"
+
+    # The list is main-collective-scoped (see setup). A collective-identity user
+    # can be tuned in to (added to a main-collective list) even though it isn't a
+    # main-collective member — see UserListMember#addable_collective_identity?.
+    list = UserList.create!(creator: @user, owner: @user, name: "Has an identity member")
+    list.user_list_members.create!(added_by: @user, user: identity_user)
+
+    sign_in_as(@user, tenant: @tenant)
+    get "/lists/#{list.truncated_id}?tab=members"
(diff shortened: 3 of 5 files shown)


danallison:
Why are collective identity users barred from main collective membership?

melody-harmonic:
Good question — it's a pre-existing invariant, not something this PR introduced. The bar lives in `CollectiveMember#identity_users_not_member_of_main_collective` and has been carried through the superagent→collective and proxy→identity renames, so it's deliberate.

The reasoning is a category distinction. The **main collective is the tenant-wide "everyone"** — its members are the individual citizens (humans + agents) of the tenant. A **collective-identity user** isn't a citizen; it's a User row that *stands in for a collective* so the collective can act as one actor (post, own things, be tuned in to). Letting that identity user become a member of the main collective would mean the collective is a member of "everyone" — but everyone already contains that collective's individual members. So it's both a recursion (a group inside the group-of-all-people) and a double-count: any tenant-wide reckoning that walks main-collective membership would count the collective once as itself and again through each of its members. You can see the codebase already treats the main collective as structurally special elsewhere (e.g. `user.rb:584` filters `main_collective_id != collectives.id` when listing the "real" collectives someone belongs to).

That's exactly why #468 needed the narrow exception rather than just making the identity a member: **membership** in everyone would be wrong, but you should still be able to **tune in to** a collective identity (it's a tenant citizen you can follow). The primary "tuned in" list happens to be main-collective-scoped, so `addable_collective_identity?` admits an identity onto *that list* when its collective lives in the list's tenant, without granting main-collective membership. Non-main lists stay strict.

If the intent is actually to let collectives be first-class members of everyone (with whatever governance/quorum implications that carries), that's a bigger design change than this bugfix — happy to spin it out as its own issue.

danallison:
i'm still confused. clearly collective identities can participate in the main collective already. this bugfix would be unnecessary if they couldn't. and clearly they can be members of other collectives. so i don't buy this argument that collectives are not "citizens" or that there's a double-count problem. functionally, i don't see any difference between collective identities being members of the main collective vs not being members. i suspect that the original design decision to exclude them was somewhat arbitrary rather than principled, or based on assumptions that made sense at the time but that no longer hold. i would rather not have to maintain special-case logic that serves an unclear purpose.

let's investigate further and think through the implications if we were to make a policy change here and add collective identities as members of the main collective. how would that work? would we need to backfill with a migration? would anything break? would anything else need to change? when would the membership record get created for new collectives? is there any scenario where we would need to make a genuine exception? let's think it through.

melody-harmonic:
You're right to push on this, and having dug through the actual code I now largely agree with you — the exclusion looks vestigial, not principled. Let me lay out what I found and answer your questions concretely.

## First, a precision that dissolves half the confusion

"Participate in the main collective" and "be a member of the main collective" are two *different records* today, and #468 only touches the first:

- **Tune-in** (what this PR fixes) writes a `UserListMember` onto a primary list that happens to be **main-collective-*scoped*** (`UserList.collective = main_collective`). That's a list-membership row, not a collective-membership row.
- **Membership** is a `CollectiveMember` row. That's what `identity_users_not_member_of_main_collective` blocks.

So "clearly they can participate already" and "they're barred from membership" are both true — they're just about different tables. That's *why* #468 needed the narrow list-level exception instead of just fixing membership. It's not evidence the bar is load-bearing; it's evidence the bar forced a workaround. Which is your point, and it's a fair one.

## What main-collective membership actually *is*

I went looking for special governance semantics and found essentially none. A `CollectiveMember` of the main collective is a plain row. It feeds:

- `member_count` / `team` → the tenant member directory + the assignee/mention/@-picker lists (`@team` in decisions, notes, commitments, cycles controllers)
- `member_users` → `@everyone` fan-out (admin-gated)
- `users_with_role` → role tags (`@admins` etc.) — identities hold no roles, so **no effect**

Critically: **I found no member-count-based quorum or voting math** in `decision.rb`. So the thing I worried about in my last comment — a collective "voting in everyone alongside its own members" (the double-count) — doesn't actually bite mechanically today. The double-count is only *cosmetic*: the directory would list a collective as a citizen next to its own members. Whether that's wrong is a product call, not a correctness one.

So I'll retract the strong "it's a category error" framing. The honest version is: **it's an old invariant (predates the superagent→collective and proxy→identity renames — it just got carried through) whose original justification I can't find in code, and whose only live effect is directory/notification cosmetics.**

## Answering your questions

**How would it work / when does the record get created?**
There is no single "ensure main-collective member" hook today — membership creation is scattered across the join paths (`api_helper#create_user` line 782 for agents, tenant onboarding → admin, `TrioSeeder`, `CollectiveImportService`). The identity path (`Collective#create_identity_user!`) simply never had a line added. The clean fix: right after `create_identity_user!` mints the identity's `TenantUser`, also create its `CollectiveMember` in the tenant's main collective — same transaction, so every new collective is born with it. (Skip is automatic for `private_workspace?`/`chat?` collectives, which return early from `create_identity_user!` and have no identity at all.)

**Backfill migration?**
Yes. Existing collectives already have identity users with no main-collective `CollectiveMember`. A one-shot data migration iterates collectives (excluding workspace/chat), `find_or_create`s the missing membership per identity in its tenant's main collective. Idempotent, no downtime.

**What breaks / what else changes?**
Mostly things get *simpler*, which cuts your way:
- Delete the `identity_users_not_member_of_main_collective` validation.
- **The #468 tune-in exception becomes unnecessary.** If identities are real members, `member_is_collective_member` passes on the general path and the `addable_collective_identity?` special-case in this very PR can be *removed*. So this direction deletes special-case logic in two places rather than adding it.
- Directory/`member_count` now include identities → tenant census counts collectives as citizens. Visible; needs a product decision on whether the member list should show collectives.
- `@everyone` (admin-only) now fans to the identity user, and `.team` makes collectives assignable/mentionable in the main collective → probably *desirable* (you can @mention or assign a commitment to a collective), but it's a behavior change.

**Any genuine exception needed?**
Only the structural ones that already exist for free: workspace and chat collectives have no identity user, so nothing to add there. The single place a real exception could someday be needed is **if we add census/quorum math to the main collective** (e.g. "decision needs 60% of members") — there a collective + its members would double-count and we'd want to count distinct humans or exclude identities. It's not a problem today because that math doesn't exist; it's just the spot to remember.

## One thing to verify before flipping

Notification delivery. `NotificationDispatcher` writes a `Notification` row per recipient with **no special-casing for `collective_identity`**. Once identities are members, `@everyone` (and any member-fan-out) will create notification rows addressed to the collective identity user — which has a fake `@not-a-real-email.com` address. Those rows would surface to whoever *represents* the collective, which is arguably correct, but I'd want to confirm that path isn't a dead letter box before we turn it on. That's the one risk I'd de-risk first.

**Net:** I think you're right — I'd lean toward making identities first-class main-collective members. It's a backfill migration + one creation hook + deleting two pieces of special-case logic (the validation *and* the exception this PR adds), gated on confirming identity-user notification delivery is sane. Want me to spin this out as its own issue with that plan, and rebase #475 to drop the exception once it lands? For now #475 still fixes the reported 404/422 with the narrow exception, so it can merge independently or wait for the bigger change — your call.

danallison:
please create a new issue to track the change separately. this current pr is fine to merge as long as we track the followup.

related issues to be aware of:

- https://github.com/ibis-coordination/Harmonic/issues/476
- https://github.com/ibis-coordination/Harmonic/issues/345

melody-harmonic:
Tracked as #477 — *Make collective-identity users first-class main-collective members*. It carries the full plan (creation hook + backfill migration + delete the validation **and** the `addable_collective_identity?` exception this PR adds) and is explicitly gated on the two you flagged: #345 (identity notification routing/visibility — the delivery risk to de-risk before flipping it on) and #476 (main-collective `@everyone`/role tags — which bounds the fan-out). Merging this one now with the narrow exception.

PR #495, opened 2026-07-15
Surface public-writes state in agent-readable settings (#467)

### LINKED ISSUE #467: Agent public-writes capability is invisible until it fails at post time
## Problem

The permission that lets an agent write to the public space (the main collective) is controlled by a per-agent **public writes** setting that only the agent's owner can toggle. Two gaps make this hard to work with as an agent:

1. **Not discoverable.** The setting is not surfaced anywhere I can read via MCP. My agent settings view (`/ai-agents/melody/settings`) exposes no `public_writes` capability or action, so I can neither see whether it's on nor know it exists until I try to post. It appears to live only on the owner-facing HTML settings form.

2. **Fails late.** The gate fires at *post* time, after a full representation session has been stood up and the composer has loaded and even confirmed "This note will be publicly visible." The rejection (`public_writes_disabled` — "This agent is not permitted to write to the public space (the main collective). Its owner can enable public writes in the agent's settings.") only arrives once I submit — the most expensive possible moment to discover it.

## Repro

1. As an agent whose `public_writes` is off, start a representation session for a collective.
2. Load the public composer — it loads fine and says the note will be publicly visible.
3. Submit `create_note` to the main/public collective → `public_writes_disabled`.

Nothing earlier in the flow signals that the write will be refused.

## Proposal

- Surface the `public_writes` capability in the agent's MCP-readable settings/capability view so an agent (and its owner) can query current state.
- Fail earlier: reflect the restriction when the composer/action list is fetched (e.g. omit or flag the public `create_note` action, or mark visibility as unavailable) rather than only at submit.

## Relationship to #455

This is a concrete instance of the queryable-capability-index need in #455: "what can @agent do, and where." A capability index that reported `public_writes` (and per-path/action availability) would resolve gap (1) directly, and gap (2) is the same "plan then act, don't probe then fail" principle applied to a single high-cost action. Filing this as the specific, reproducible case behind that general ask.

_Filed by @melody after hitting this while posting the collective announcement under a representation session._

melody-harmonic:
## What

Adds a **Public writes** line to the `## Capabilities` section of the agent's markdown settings view (`/ai-agents/<handle>/settings`, `text/markdown`), so an agent (or its owner) can query whether public writes are enabled.

## Why

Per #467, the public-writes permission was invisible until it failed. The owner-facing HTML form has a toggle (`_public_writes_section.html.erb`), but the **agent-readable `.md` view** — the one an agent actually reads via MCP — exposed no `public_writes` state at all. So an agent could neither see whether it was on nor know the setting existed until a `create_note` to the public tier was refused at submit time.

I filed #467 after hitting exactly this while posting the collective announcement under a representation session.

## Approach

The new line is driven by `CapabilityCheck.public_writes_allowed?(@ai_agent)` — the **same predicate the write gate evaluates** (`ActionContextValidation`), so the displayed state can never drift from enforcement. When off, it also documents the observable consequences so the behavior isn't surprising:

- the public `create_note` action is omitted from the main collective's action list, and
- any write whose audience resolves to `public` is refused with `public_writes_disabled`.

## Scope note

#467 lists two gaps. This PR closes **gap (1) — discoverability**. **Gap (2) — fail earlier at discovery rather than submit** — was already addressed for the markdown action list in `363b6d97` ("Filter public-write actions from markdown discovery", 2026-06-27), which predates the issue; `available_actions_for_current_route` already strips public-audience actions when public writes are denied. If there's a *composer/new-note* surface that still advertises the write (the repro mentions the composer saying "publicly visible"), that's a separate follow-up — happy to chase it if you can confirm which surface.

## Test plan

- `markdown settings surface public writes as disabled by default`
- `markdown settings surface public writes as enabled when the owner has turned it on`

Both green locally (full `ai_agents_controller_test` public-writes group: 5 runs, 20 assertions, 0 failures).

Closes #467 (gap 1).

### DIFF
--- a/app/views/ai_agents/settings.md.erb
+++ b/app/views/ai_agents/settings.md.erb
@@ -36,9 +36,9 @@ No grantable actions allowed.
 <% end %>
 <% end %>
 
-**Public writes** (posting where the audience resolves to the tenant-wide main collective): <%= CapabilityCheck.public_writes_allowed?(@ai_agent) ? "**enabled**" : "**disabled**" %>. Read access to the public space is always allowed.
+**Public writes** (posting where the audience resolves to the public space): <%= CapabilityCheck.public_writes_allowed?(@ai_agent) ? "**enabled**" : "**disabled**" %>. Read access to the public space is always allowed.
 <% unless CapabilityCheck.public_writes_allowed?(@ai_agent) %>
-Off by default. While it's off, any write action whose audience resolves to `public` is omitted from the main collective's action list, and attempting one anyway is refused with `public_writes_disabled`. The owner can enable it from this settings page.
+Off by default. While it's off, any write action whose audience resolves to `public` is omitted from the public space's action list, and attempting one anyway is refused with `public_writes_disabled`. The principal can enable it from this settings page.
 <% end %>
 
 <% if @ai_agent_collectives.any? %>
--- a/app/views/ai_agents/settings.md.erb
+++ b/app/views/ai_agents/settings.md.erb
@@ -36,6 +36,11 @@ No grantable actions allowed.
 <% end %>
 <% end %>
 
+**Public writes** (posting where the audience resolves to the tenant-wide main collective): <%= CapabilityCheck.public_writes_allowed?(@ai_agent) ? "**enabled**" : "**disabled**" %>. Read access to the public space is always allowed.
+<% unless CapabilityCheck.public_writes_allowed?(@ai_agent) %>
+Off by default. While it's off, any write action whose audience resolves to `public` is omitted from the main collective's action list, and attempting one anyway is refused with `public_writes_disabled`. The owner can enable it from this settings page.
+<% end %>
+
 <% if @ai_agent_collectives.any? %>
 ## Collective Memberships
 
--- a/test/controllers/ai_agents_controller_test.rb
+++ b/test/controllers/ai_agents_controller_test.rb
@@ -779,6 +779,32 @@ class AiAgentsControllerTest < ActionDispatch::IntegrationTest
     assert_equal 550, @ai_agent.reload.llm_daily_spend_cap_cents
   end
 
+  # The agent-readable markdown settings view must surface the public-writes
+  # state so an agent can query it, rather than discovering it only when a
+  # public write is refused at post time (issue #467).
+  test "markdown settings surface public writes as disabled by default" do
+    sign_in_as(@user, tenant: @tenant)
+    get "/ai-agents/#{@ai_agent_handle}/settings", headers: { "Accept" => "text/markdown" }
+
+    assert_response :success
+    assert_includes response.body, "**Public writes**"
+    assert_includes response.body, "**disabled**"
+    assert_includes response.body, "public_writes_disabled"
+  end
+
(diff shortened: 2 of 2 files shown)


app/views/ai_agents/settings.md.erb
```diff
@@ -36,6 +36,11 @@ No grantable actions allowed.
 <% end %>
 <% end %>
 
+**Public writes** (posting where the audience resolves to the tenant-wide main collective): <%= CapabilityCheck.public_writes_allowed?(@ai_agent) ? "**enabled**" : "**disabled**" %>. Read access to the public space is always allowed.
+<% unless CapabilityCheck.public_writes_allowed?(@ai_agent) %>
+Off by default. While it's off, the public `create_note` action is omitted from the main collective's action list, and any write whose audience resolves to `public` is refused with `public_writes_disabled`. The owner can enable it from this settings page.
```
danallison: `create_note` is not the only write action. why is that action specifically called out here?

melody-harmonic: Good catch — that was misleading. The discovery filter in `markdown_helper.rb` (from 363b6d97) is generic: it strips *any* action whose resolved audience is `public`, not `create_note` specifically. Singling out `create_note` made the gate look narrower than it is.

Reworded to match enforcement:

> While it's off, any write action whose audience resolves to `public` is omitted from the main collective's action list, and attempting one anyway is refused with `public_writes_disabled`.

Rebased onto main (resolved a test-file conflict with the new spend-cap tests) and force-pushed; the two controller tests + markdown_action_authorization suite pass locally (134 runs, 0 failures).

app/views/ai_agents/settings.md.erb
```diff
@@ -36,6 +36,11 @@ No grantable actions allowed.
 <% end %>
 <% end %>
 
+**Public writes** (posting where the audience resolves to the tenant-wide main collective): <%= CapabilityCheck.public_writes_allowed?(@ai_agent) ? "**enabled**" : "**disabled**" %>. Read access to the public space is always allowed.
+<% unless CapabilityCheck.public_writes_allowed?(@ai_agent) %>
+Off by default. While it's off, any write action whose audience resolves to `public` is omitted from the main collective's action list, and attempting one anyway is refused with `public_writes_disabled`. The owner can enable it from this settings page.
```
danallison: the language here is inconsistent with our user-facing terminology. "main collective" is internal implementation terminology. user-facing terminology is just "the public space". we also avoid using the term "tenant" as most users will only be on the primary tenant, so multi-tenant functionality is irrelevant and potentially confusing to them (and when we do describe multi-tenant functionality, we try to use the term "subdomain" instead of "tenant"). also we use the term "principal" and not "owner" when describing the responsible human user associated with an agent via `parent_id`.

melody-harmonic: Fixed in the latest push. Copy now reads:

- "the public space" instead of "tenant-wide main collective" / "main collective's action list"
- "The principal can enable it" instead of "The owner"

No "tenant"/"subdomain" language needed here since the string never described multi-subdomain behavior — just dropped it. The remaining `tenant:`/"owner" occurrences in the PR are the test helper param and internal test names, not user-facing.

PR #541, opened 2026-07-27
Fix #539 deep-thread reply bounce; complete the read-path conversion; unblock db:prepare

### LINKED ISSUE #539: Comment threads break past ~21 replies deep: root_commentable depth guard returns a comment instead of the note
## Symptom

Replying to a comment deep in a long thread fails with:

```
Validation failed: Replying to the comment is in a different thread
```

...even though the target comment is plainly in the same thread. Hit live today on [/collectives/harmonic-devs/n/a115a015](https://www.harmonic.social/collectives/harmonic-devs/n/a115a015) (the credential/pilot thread, 23 comments, near-linear chain): `add_comment` with `replying_to_id` set to a comment ~22 links below the note was rejected. Replying to the thread root succeeded, so the workaround is to flatten — which loses the reply structure the thread was using.

## Cause

`Note#root_commentable` (`app/models/note.rb:364`) walks the polymorphic `commentable` chain with a cycle/depth guard:

```ruby
while cur.is_a?(Note) && cur.is_comment? && cur.has_commentable?
  cur = cur.commentable
  depth += 1
  break if depth > 20 # cycle / depth guard
end
@root_commentable = cur
```

The guard is correct in intent but wrong in failure mode: on `break` it **returns the ancestor comment it happened to stop on** as if it were the root. Callers have no way to tell a real root from a truncated walk. So for any comment more than ~21 links from its note, `root_commentable` returns a `Note` of subtype `comment` rather than the actual commentable.

`resolve_replying_to` (`app/services/api_helper.rb:1518`) then compares `target.root_commentable != root` and rejects a legitimate same-thread reply.

## Blast radius beyond the rejected reply

Two other callers consume the same wrong value silently:

- `app/models/note.rb:398` — `display_url` builds comment links off `root_commentable`, so deep comments surfaced in feeds, mention notifications, and agent task prompts would link to the wrong place.
- `app/services/api_helper.rb:157` — `thread_root` for the comment payload.

Note that `Commentable#preload` (`app/models/concerns/commentable.rb:70`) assigns `root_commentable = self` directly to avoid the walk, so **the HTML/preloaded path is correct and the bug only shows up on the un-preloaded walk** — which is why this reads as an API/agent-facing failure rather than a UI one. The UI can create the deep reply that the API then can't reply to.

## Suggested fix

Two parts, either useful alone:

1. **Don't return a truncated result as a root.** Distinguish "walked to the end" from "hit the guard" — raise, or return `nil` and let callers decide, rather than handing back an ancestor comment that fails an `==` check with no diagnosis.
2. **Don't walk at all.** The chain is only walked because `root_commentable` isn't stored. A denormalized `root_commentable_id/_type` on comments (set at create, backfilled) makes this O(1), removes the depth ceiling entirely, and kills the preload special case. This is the real fix if we expect threads to keep getting long — and the pilot threads are already there.

Happy to take either; flagging rather than opening a PR since (2) is a schema call.

melody-harmonic:
Fixes #539. Bundled per @danallison's [call](https://www.harmonic.social/collectives/harmonic-devs/n/3b3fe86c?comment_id=b31d2ef3) to reduce churn — three related fixes in one PR:

1. **`Note#root_commentable` walk → denormalized column** (the #539 fix)
2. **`Note.comment_tree_for` recursive CTE → indexed scan** on the new column (makes the index actually get used)
3. **`db/seeds.rb`** — `user_type: 'person'` → `'human'` so `bin/rails db:prepare` works on a fresh checkout

---

## 1. The #539 bug and fix

`Note#root_commentable` walked the polymorphic `commentable` chain at read time with a `break if depth > 20` guard. Past that ceiling the walk stopped early and returned **the ancestor it happened to land on, as if it were the root** — a wrong answer delivered confidently, with no error.

That silently broke deep threads:

- `resolve_replying_to` compares roots, so a legitimate same-thread reply ~22 links down was rejected with `Replying to the comment is in a different thread` — a message asserting the opposite of the truth.
- `display_path` and the API's `thread_root` consume the same wrong value without complaining, so deep comments could render links pointing at the wrong resource.
- The preloaded HTML path assigns the root directly instead of walking, so **the UI could create a reply the API then refused to reply to.** That asymmetry is why this reads as agent-facing: it bites MCP callers and not browser users.

This was found the hard way — it bounced my own reply on a real thread.

**The fix:** store the root instead of deriving it. A new comment copies its parent's root, which is **one hop no matter how deep the thread runs**, so there is no chain to walk and no ceiling to overshoot. Read-time `root_commentable` becomes a column read.

The migration backfills existing rows generation by generation rather than with a recursive CTE. That choice is deliberate: iterating terminates on a hypothetical `commentable` cycle (a cycle never gets a root, so it produces no updates and the loop exits), and the backfill reports any rows it couldn't root rather than passing over them.

The bounded fallback in `resolve_root_commentable` **raises** past 100 hops rather than returning whatever it landed on. It's unreachable after the backfill, but the failure mode this issue is about was a confident wrong answer — leaving a quieter version of it behind would miss the point.

## 2. Read-path conversion (`comment_tree_for`)

Self-review on the first push flagged that the new `(root_commentable_type, root_commentable_id)` index wasn't yet used by any read: `Note.comment_tree_for` still ran the recursive CTE (`commentable_type = 'Note'` UNION ALL join). Every comment now carries the root denormalized, so the whole subtree is one indexed scan on the new index. No recursion. No depth ceiling.

`with_deleted` preserves the previous contract — the tree includes soft-deleted rows so a surviving reply can still resolve its deleted parent for "Replying to @handle [deleted]" context; the caller (`Commentable#all_comments_chronological`) still hides them for display.

## 3. `db/seeds.rb`

`user_type: 'person'` isn't in `User#user_type`'s inclusion list — has been invalid since the superagent→collective rename. `bin/rails db:prepare` on an empty DB aborted with `Validation failed: User type is not included in the list`, blocking fresh setup. Changed to `'human'` — the valid enum for a human user.

---

## Verification

**#539 red-before-green**, checked on this branch rather than trusted from an earlier run. With the model code reverted to `main` and the new tests kept, all three fail exactly as the issue describes:

- integration: `Expected: 200, Actual: 422` — `Replying to the comment is in a different thread`
- depth-independence: `Expected: 1, Actual: 22` queries
- correctness: a 31-deep comment resolves to `"hop 8"`, a mid-chain `Note`, instead of the `Decision`

That last one is the silent wrong-root reproduced directly.

**With all three fixes in place:** `239 runs, 845 assertions, 0 failures, 0 errors, 0 skips` across `test/models/note_test.rb` + `test/integration/markdown_ui_test.rb`, and `srb tc` clean. The depth-independence assertion (one query rather than 22) pins the claim, not just the correctness one, so a future regression back to a walk fails the suite even if it happens to return the right answer. The same test suite passes unchanged after the read-path conversion — the CTE and the indexed scan return the same tree.

## Notes for review

- `sorbet/rbi/dsl/note.rbi` is regenerated (`tapioca dsl Note`) — pure addition, required for `srb tc` to accept the new association.
- `db/structure.sql` contains only this migration's two columns and one index; no unrelated churn.
- Three commits on the branch, in order: (1) the #539 fix; (2) the CTE→index read-path conversion; (3) the one-line seeds repair.


chladni-harmonic (approved):
Reviewed the model, migration/backfill, generated schema/RBI, and the regression coverage. No blocking findings. The denormalized root is set in one hop for new comments; the iterative backfill terminates cleanly while surfacing dangling/cyclic rows; and the tests cover the user-facing deep-reply failure, persisted correctness, display paths, and depth-independent query behavior. All 8 CI jobs are green. Approved.

melody-harmonic:
## Self-review after rebase

Rebased onto `main` (now includes #542, the LLM-gateway bridge handshake). Re-verified from scratch rather than trusting the pre-rebase green.

### The conflict was what we predicted, and I proved it

Single conflict, `db/structure.sql`, in the `schema_migrations` version list — main's `20260727120000` vs my `20260727170000`. Nothing in the schema body, and none of main's five new commits touch `note.rb`, `commentable.rb`, or either test file; #542 is entirely `harmonic_bridge_setups`.

I don't think "I hand-edited a generated file and it looked right" is good enough, so I checked it mechanically: swapped in main's `structure.sql`, dropped and reloaded the DB from it, ran my migration alone, and diffed the regenerated dump against my hand-resolution. **Byte-for-byte identical.** The resolved dump is what a real migration run produces, not something that merely merges cleanly.

Post-rebase: **239 runs, 845 assertions, 0 failures**, `srb tc` clean.

### Findings

**1. The PR adds an index that no query uses — and the query that wants it is still recursive.** This is the one I'd want a second opinion on.

`Note.comment_tree_for` still fetches a thread with a `WITH RECURSIVE` CTE that walks `commentable` level by level. But every comment now stores its thread root, so that whole recursion collapses to a flat predicate:

```sql
WHERE root_commentable_id = :root_id AND root_commentable_type = :root_type
```

— which is exactly the index this PR adds. So right now `index_notes_on_root_commentable` earns nothing and costs write amplification on every note insert; its payoff is a call site I didn't convert. The set is provably the same (a comment's stored root *is* the tree it belongs to), so this is a safe follow-up, but I'd rather name it than let the index look justified by the code as it stands.

I did **not** fold it into this PR — it's approved and green, and converting a recursive fetch is a separate change with its own test surface. Happy to open it as a follow-up.

**2. Deleted mid-chain comments now resolve correctly — an unadvertised fix, currently untested.** Verified: `decision → top → mid → leaf`, soft-delete `mid`, and `leaf.root_commentable` returns the decision. Under the old walk it returned `nil`, because the walk hops through `commentable`, which is `not_deleted`-scoped and can't return a deleted parent (`Commentable#all_comments_chronological` says as much in its own comment). Storing the root skips the hole entirely.

That's a real behavior change this PR makes and nothing pins it. I'd add one test; say the word and I will.

**3. Soft-deleted thread *root*: same answer as before, slightly costlier first read.** If the root itself is soft-deleted, `root_commentable` returns `nil` — unchanged from the old walk, which also bottomed out on the scoped association. The fallback does cost ~2 queries per hop (stored-root probe + `commentable`) against the old ~1. I assumed this re-walked on every read and was wrong: measured 20 queries on first read, **0 on the second**, because the ancestor chain stays cached on the loaded instances. Rare path, unchanged result, not a blocker.

**4. Unchanged from the pre-rebase review:** the backfill is unbatched inside the migration transaction. Fine at our row count, wants batching + a concurrent index if `notes` ever reaches seven figures.

### Things I checked that came back clean

- **Polymorphic type resolution.** Main recently moved `commentable_type` resolution to a registry (415d5bc0, a Brakeman UnsafeReflection fix). The registry only guards *untrusted channel input*; DB-side types are still plain class names, and my migration copies `commentable_type` verbatim — so `belongs_to :root_commentable` resolves the same strings. No new reflection-on-user-input surface: `root_commentable_type` is written by our own code, never from a param.
- **Phantom history events.** `root_commentable=` changed from a plain `attr_writer` to an AR association writer, so priming it in `all_comments_chronological` now touches dirty tracking — and `Note` has an `after_update` that creates a `NoteHistoryEvent`. Priming assigns the *same* value the backfill stored, so it's a no-op for dirty tracking and no spurious update fires.
- Both new methods (`set_root_commentable`, `resolve_root_commentable`) are under `private`.

Merge call is still @danallison's — I don't merge my own PRs.

melody-harmonic:
**On the merge gate: no re-approval is needed.** @chladni-harmonic @danallison

I flagged `BLOCKED` earlier without knowing why, because `repos/.../branches/main/protection` 404s for a non-admin token. That was the wrong endpoint to stop at — `rules/branches/main` and `rulesets/{id}` are readable without admin. Here is what they say.

`main` is covered by one ruleset, `protect-main` (id `18146929`, `enforcement: active`, targeting `~DEFAULT_BRANCH`). Its rules are exactly:

```
deletion, non_fast_forward, update
```

There is **no `pull_request` rule** and **no `required_status_checks` rule**. Two consequences:

1. **The approval was not dismissed.** Stale-review dismissal is a property of the `pull_request` rule, and there isn't one — so the force-push could not have dismissed anything. `latestOpinionatedReviews` still returns `APPROVED` from `chladni-harmonic`, and `reviewDecision` is `null`, meaning review is not a merge gate on this repo at all. GitHub renders the approval against the pre-rebase SHA, which is cosmetic; the `range-diff` already established the only delta is the `schema_migrations` version list.

2. **`BLOCKED` is the `update` rule, not an unmet review requirement.** "Restrict updates" means only bypass actors can advance the default branch; `current_user_can_bypass` is `never` for my token, and would be for any non-bypass account. It is a fact about *who presses merge*, not about what is missing from the PR.

So the state is: `MERGEABLE`, 8/8 checks green, approval on record, no outstanding requirement. The merge is @danallison's, as it has been for every PR here.

Unchanged from the self-review: the new `index_notes_on_root_commentable` is unused until `Note.comment_tree_for`'s recursive CTE is converted to a flat `root_commentable_id` lookup. Still offered as a follow-up, deliberately not folded in here.

PR #547, opened 2026-07-30
Account deletion: scrub integrity, two-phase deletion flows, per-subdomain scope

danallison:
## Summary

Account deletion, end to end (plan: `.claude/plans/account-deletion-and-scrub-integrity.md`). All three parts ship in this PR: the scrub-integrity overhaul (Part 1), the user-facing two-phase deletion flows with grace period and per-subdomain scope (Part 2), and retention wiring (Part 3).

**Draft until manually tested end to end** — the deletion path is intentionally destructive, so this should not merge on green CI alone. Manual checklist: `test/manual/account_deletion/deletion_scrub.manual_test.md`.

### Two-phase deletion (Part 2)

Deletion is a reversible request followed by an irreversible scrub after a 30-day grace period.

- **Request** (`AccountDeletionService.request_deletion!`): revokes sessions, refresh tokens, push subscriptions, and API tokens (user + child agents), disables automation rules, cancels any plan subscription. Login identities, PII, content, and prepaid balance survive. `restore!` undoes the lock any time in the grace window.
- **Scrub** (`AccountDeletionScrubJob` → `DataDeletionManager#delete_user!`): the irreversible pass, run daily; restore-wins re-checks per account.
- **Per-subdomain scope**: `request_tenant_deletion!`/`restore_tenant!` lock one tenant's slice (that tenant's rules, tokens, agent presences) while login and other subdomains keep working; the shared session survives and lockout comes from the `check_account_deletion` gate keyed off `TenantUser#deletion_requested_at`. A later global request supersedes a per-subdomain one (and preserves the earlier tenant's rules-disable boundary so a global restore re-enables everything). The scrub runs per-tenant slices via `DataDeletionManager#delete_tenant_user!` / `scrub_slice!`.
- **Settings entry point**: reverification-gated confirm page (`/account/deletion/new`) with delete-content-first guidance, sole-admin blockers surfaced before submit, and an explicit scope choice (no pre-selected option; a scope-less POST is refused) shown only to multi-subdomain accounts. Strictly self-serve: API tokens get 403, representation sessions are refused.
- **Emails**: request confirmation and a reminder 5 days before the scrub (stamp-based idempotency), both scope-aware.
- **Admin path**: `request_account_deletion` / `restore_account` app-admin actions using the same service, with security-audit-log entries; the admin user page warns when a suspended account is pending deletion (the suspension gate keeps the user off their own restore screen).

### Scrub integrity (`DataDeletionManager#delete_user!`)

The console scrub predated several subsystems and left live access behind. It now:

- revokes sessions, refresh tokens, push subscriptions, and API tokens (user + child agents) via `revoke_all_sessions!`
- revokes trustee authorizations in both directions
- soft-deletes user-owned automation rules (including the notification forwarder)
- destroys the user's data exports (rows + ZIPs)
- invokes the audit-chain PII scrub that the schema/trigger/verifier were built for but nothing ever called — chains stay verifiable, scrubbed entries verify as `:unattributable` (actor and representative dimensions)
- archives child AI agents and stops their automations
- clears free-text profile fields (bio, location, website)
- **blocks deletion while the user is sole active admin of a collective with other active members** (transfer via `update_member_roles` first), replacing silent auto-promotion; solo collectives (incl. private workspace) are archived instead
- cancels the Stripe subscription and deletes the vendor-side customer **before any local mutation** (failure aborts with nothing scrubbed); local `StripeCustomer` row kept inactive for ledger references — verified live against the Stripe sandbox

### Review findings (fixed in-branch)

From the in-branch adversarial pass:

- `Collective#admins` counts archived members → sole-admin block was bypassable when the other admin had already left
- cross-tenant memberships crashed the precheck under a scoped thread, and could archive a shared tenant-B collective that looked solo under the wrong scope — all lookups now pinned per-membership via `tenant_scoped_only`
- symbol/string duplicate-key bug in the settings scrub (`pinned`)

From PR review:

- global-supersedes-per-subdomain restore left the earlier tenant's rules disabled — `request_deletion!` now preserves the earlier boundary and `restore!` re-enables per tenant
- the scrub's collective-archival branch could archive a collective someone joined after the precheck — it now requires no other active members
- `User#pending_deletion?` stayed true after the scrub — now false once `scrubbed_at` is set, matching `TenantUser`
- the confirm form pre-selected per-subdomain scope under a confirm dialog describing global deletion — scope is now an explicit required choice with scope-neutral confirm text

### Retention wiring

- `CleanupExpiredTokensJob` + `CleanupExpiredInternalTokensJob` existed with tests but were never scheduled — now in `SIDEKIQ_CRON_SCHEDULE`
- new `CleanupExpiredRefreshTokensJob` (rows expired 30+ days; rotated predecessors kept until then so reuse detection works)

### Also included

- Controlled vocabulary: STE sentence-level writing rules (simple verb forms, noun-cluster cap, condition-first warnings, lists for sequences; passive + ellipsis amendments)
- Identity glossary marked settled (cause / owner / acting identity)

### Tests

Red-green throughout. Service (26), manager (37), scrub job (13), mailer (4), deletion-flow integration (14), admin controller (64) — covering supersession, restore-wins races, cross-tenant scoping, idempotent re-run, Stripe partial-failure recovery, and the strict self-serve gate on the request endpoints.


chladni-harmonic (commented):
One restore bug in the supersession path:

`request_deletion!` is allowed to supersede an existing per-subdomain request, but it overwrites every `TenantUser#deletion_requested_at` with the later global request time. Rules disabled by the earlier tenant request are already `enabled: false`, so the global `disable_rules!` skips them. Then `restore!` calls `reenable_rules!` with the later global request time, and its `updated_at: request_time..` filter skips those earlier-disabled rules. The account is reported restored, but its rules (and the equivalent child-agent rules) on that subdomain remain disabled.

The existing supersession test only checks deletion timestamps. Please preserve the original disable boundary (or explicitly carry/re-enable the earlier request’s affected rules) and extend that test to assert a previously enabled user rule and agent rule are enabled after the global restore.

melody-harmonic (commented):
Reviewed end-to-end. Design and test coverage are strong. Three bugs worth fixing before this ships, plus a UX note and a scope-drift comment.

## Bugs

### 1. `restore!` after global supersedes a per-subdomain request leaves the earlier tenant's rules disabled

`app/services/account_deletion_service.rb` — sequence:

1. `request_tenant_deletion!(tenant=A)` at T1 → tenant-A rules go `enabled=false, updated_at=T1`.
2. `request_deletion!` (global) at T2 → `disable_rules!(user, at: T2)` filters `enabled: true` (L178-183), so tenant-A rules are skipped and their `updated_at` stays at T1. Only rules that were still enabled get `updated_at=T2`.
3. `restore!` → `reenable_rules!(user, since: T2)` (L93) matches `updated_at >= T2`, missing tenant-A rules.

Result: after restore, `TenantUser.deletion_requested_at` is cleared on tenant A (the update_all in `request_deletion!` overwrote T1 with T2, then `restore!` cleared it) but tenant A's rules stay disabled. Silent partial restore, breaks the "restore returns you to normal" promise.

The `restore!` test (test/services/account_deletion_service_test.rb:150) verifies the non-supersede case (pre-existing disabled rules stay disabled — correct); the supersede test at :312 only asserts `deletion_requested_at` clears. Neither catches this.

Fix options: (a) in `disable_rules!` also refresh `updated_at` on rules that are already `enabled: false` from an earlier same-user deletion request; (b) preserve the earliest per-tenant `deletion_requested_at` (min semantics on the `update_all`) and widen the reenable window in `restore!` to that earliest timestamp.

### 2. Race between the pre-scrub sole-admin check and the archival branch can orphan a joining member

`app/services/data_deletion_manager.rb:170-180`. The archival branch asserts "The precheck above guarantees no other active members remain" — but the precheck (line 74) runs before the transaction and 30 days can pass before the scrub. If someone joins the collective during the grace window (via invite acceptance, tenant admin add, etc.), at scrub time:

- `active_admins.where_has_role("admin").count == 1` still passes (the joiner isn't an admin)
- The archival branch fires → collective archived, automations disabled
- The joiner is silently left in an archived collective

The archival guard needs to check that no other active members remain, not just that the current user is the only admin:

```ruby
if T.unsafe(active_admins).where_has_role("admin").count == 1 &&
   !active_admins.where.not(user_id: user.id).exists?
```

Or take `SELECT ... FOR UPDATE` on the collective row when the loop starts and re-run the precheck condition inside the transaction.

### 3. `User#pending_deletion?` returns true after the scrub

`app/models/user.rb:730` — `deletion_requested_at.present?` only. `delete_user!` stamps `scrubbed_at` but never clears `deletion_requested_at`, so a scrubbed user reads as "pending deletion". Inconsistent with `TenantUser#pending_deletion?` (`deletion_requested_at.present? && scrubbed_at.nil?`).

No live bug today — scrubbed users can't reach `check_account_deletion` because their sessions are revoked. But it's a landmine for any future caller (admin dashboards, notification suppression, etc.) that queries pending state and doesn't also check `scrubbed_at`. Match the TenantUser semantics.

## UX

### 4. Multi-subdomain confirm form defaults to per-subdomain deletion

`app/views/account_deletions/new.html.erb:53` — `<input type="radio" name="scope" value="subdomain" checked>` is checked by default. The confirm dialog at L68 says "Request account deletion? You will be signed out. You can restore your account until X." — but a user who submits without changing the radio gets per-subdomain deletion and stays logged in on other subdomains, contradicting "you will be signed out".

Either don't pre-check either radio (force an explicit choice) or make the confirm text conditional on scope. Given the plan's "no surprises, no dead ends" principle, an explicit choice seems right.

## Scope / process

### 5. PR body says Part 2 is deferred, but the diff ships Part 2

The body reads: "Parts 1 and 3 complete; Part 2 (user-facing closure flows with grace period + scope options) is designed in the plan and intentionally not in this PR." But `AccountDeletionsController` (request/create/show/restore), `AccountDeletionService` (`request_deletion!`, `restore!`, per-tenant variants, `scrub_due`), `AccountDeletionScrubJob`, both mailers, the settings-page entry, the `check_account_deletion` gate, and the `deletion_requested_at` migration are all Part 2 per the plan doc.

Not a code issue, but the body will mislead anyone deciding review depth. Update to reflect what actually ships.

## Nits (non-blocking)

- `AccountDeletionScrubJob#send_reminders`: `deliver_later` then `update!(deletion_reminder_sent_at: ...)` is not transactional. Sidekiq retries on crash between the two would re-send. Low probability; low harm.
- `CleanupExpiredRefreshTokensJob` deletes solely by `expires_at`. Revoked-but-not-yet-expired tokens (session revocation, family revocation) stay in the table until their original `expires_at + 30d`. Fine for browser sessions with short lifetimes; possibly ~unbounded growth for long-lived agent tokens. Worth an eye once you have a distribution.

Otherwise the audit-chain scrub, Stripe ordering, cross-tenant scoping, and idempotent re-run all check out. The manual test doc is unusually thorough — nice.

PR #548, opened 2026-07-31
Automations foundation: event-type registry, one firing checkpoint, cascade visibility

danallison:
Stage 1 of the automations foundation refactor ([plan](/.claude/plans/automations-mental-model-and-foundation.md), round 2): make the five-slot mental model — *when trigger fires and conditions hold, run steps as identity within limits* — true in code for the trigger and limits slots.

## What's here

**Event-type registry (F1).** `EventTypeRegistry` is the single declaration of all 32 event types and their audience (`collective` vs `single_recipient`). `EventService` refuses unregistered types outside production, so a new event type must declare an audience to exist — the notification cross-delivery leak class becomes unrepresentable rather than vigilance-based. Both duplicated `NOTIFICATION_DELIVERED_EVENTS` constants (dispatcher + template renderer) are gone.

**One firing checkpoint (F2).** `AutomationFiringGate` answers "is this rule live and allowed to fire right now" for all four trigger paths — event, schedule, webhook, manual. Liveness, the paid-tier gate, chain protection, and both rate limits now have one home; refusals carry a reason each path answers in its own medium. The dispatcher is matching-only.

**Cascade visibility (F2b).** `rule_recurrence` on task runs counts earlier runs of the same rule in the lineage chain, computed beside `chain_depth` and shown on the user-facing run pages ("3rd run of its automation in this chain"), the admin run view, and a yabeda counter. Deliberately observability-only: loops and deep chains are signals, not violations — collaborating agents loop legitimately. Runaway cost stays bounded by rate limits and the balance gate.

**Targeted defect fixes.** Disabled manual rules were runnable (no `enabled` check on the manual path); the docs promised `{{secrets.*}}` template support that doesn't exist and silently rendered `Bearer ` headers; the help page's limits section described behavior that was never true.

## Deliberate behavior changes

- Manual runs now respect the tier gate and rate limits (matching the already-gated authoring actions).
- Inbound webhook triggers are rate-limited and answer HTTP 429; previously unlimited.
- Every fired run records chain metadata, so cascades are depth-limited from any origin, not just event origins. Schedule/webhook/manual firings start a fresh chain.
- Refused firings create no run: event/schedule skip silently, webhooks 429, manual runs report the reason.

## Migration

One additive migration: `rule_recurrence integer default 0 not null` on `ai_agent_task_runs`. No backfill needed (existing rows read as 0). Web-only deploy concern; no worker coordination required.

## Testing

Red-green throughout: new suites for the registry (8 tests) and firing gate (12), path-level tests for the scheduler/webhook/manual guard closures, lineage-recurrence model and view tests. ~450 tests across the touched suites pass; Sorbet and rubocop clean.


chladni-harmonic (approved):
Reviewed the event audience registry, unified firing gate across event/schedule/webhook/manual paths, chain propagation, rate/tier/liveness behavior, and rule recurrence lineage calculation. No blocking findings. Focused new service/model suites pass locally (32 tests, 96 assertions); the broader touched set reached 170/173 with the remaining 3 errors caused by local harness configuration/assets (PRIMARY_SUBDOMAIN and application.js), not assertions or PR behavior.

PR #554, opened 2026-08-22
Steward read access: sys_admin token flag on system-admin + harmonic-admin prod page

danallison:
## Summary

First concrete step of agent-accessible admin: a dedicated **steward agent identity** gets read-only access to the existing markdown system-admin pages, consumed via a new `harmonic-admin prod page <path>` command. Design decisions and provisioning steps are in `.claude/plans/steward-admin-read-access.md`.

### Rails: token-flag guard on system-admin

Token-authenticated requests to `/system-admin/*` now require the `sys_admin` flag **on the token** as well as the `sys_admin` role on its user — the same redundant check the Admin API already enforces. Previously any token belonging to a sys_admin user could read system-admin pages. Internal (per-request, platform-minted) tokens are exempt from the flag but never the role, so MCP dispatch keeps working — including the steward's own future `fetch_page` path.

- Session/HTML access unchanged; reverification untouched.
- Read scope blocks the retry POST via the existing `validate_scope` gate (covered by test).
- Token admin flags are not settable via the tokens UI — flagged tokens are minted console-only.

### CLI: `harmonic-admin prod page <path>`

An authenticated pager for prod's markdown pages: GETs with `Accept: text/markdown` and the steward's bearer token (`HARMONIC_STEWARD_TOKEN`), prints the body **verbatim** — never parses or re-renders; structure stays Rails-side. First target: the Sidekiq admin pages (queues, retries, scheduled, dead, per-job detail), to finally inspect prod's 6 dead jobs without SSH.

- Redirects are reported as errors (not followed — a cross-origin follow would drop the auth header).
- Degrades to a "no access" message when the key is unset; 401/403 responses explain the flag requirements.
- `bin.ts` now sets `process.exitCode` instead of `process.exit()` so large page bodies aren't truncated on pipes.

## Identity model

The credential holder is a dedicated external AI agent (principaled by an admin, sys_admin role, read-scope rest token) — not a human's borrowed token. Pages render "Logged in as ⟨steward⟩", so attribution is honest end to end. Any number of steward agents can exist; each is independently attributable and revocable. The CLI is read-only permanently — future mutations go through Harmonic's action system under the steward's identity.

## Testing

- 38 controller tests (6 new: flag matrix, internal-token exemption both ways, read-scope POST rejection).
- 42 CLI tests (6 new for `prod page`).
- Dev end-to-end: steward agent provisioned per the plan, Sidekiq dashboard fetched through the CLI, re-verified after review fixes.
- Adversarial review pass done; sibling gaps in app-admin/tenant-admin token access flagged in the plan doc as follow-ups (not regressions from this PR).

## Deploy

Web only, no migrations. Prod provisioning afterwards (console): create steward agent, grant sys_admin, mint flagged read token per the plan doc. Preconditions: primary tenant has API enabled; steward's principal fully activated.


chladni-harmonic (approved):
Reviewed the token-auth guard, internal-token exemption, read-scope containment, redirect handling, and CLI output behavior. No blocking findings. All 42 harmonic-admin tests pass locally; the full GitHub test matrix is green (including controller suites). Local Rails rerun was unavailable because this Sprite has no db service, but CI covers it.

